PR #5, opened 2014-11-19
remove GNU-specific usage for `sed` in shell example

ericvw:
This ensures that the shell example works with GNU and non-GNU `sed`.


ericvw:
@tmikov, can you ensure this usage of `sed` works for your non-GNU version?

PR #6, opened 2014-12-22
BLPAPI wrapper and miscellaneous improvements

ericvw:
These set of commits primarily refactored and improved the internal details of `blpapi-wrapper.js`.  The major takeaways are:
1. Properly handling all session termination in the event of asynchronous events
2. Improving the usage of TypeScript by adding precise type definitions
3. Reduce the usage of internal state that must be kept by re-using the already uniquely generated correlation ids
4. Extensive and proper usage of logging has been utilized

The other miscellaneous improvement has been improving the the `example/ex1.sh` script to echo a newline at the end to ensure the command prompt would be flushed left after execution.

### DIFF
--- a/tslib/blpapi-wrapper.ts
+++ b/tslib/blpapi-wrapper.ts
@@ -3,7 +3,7 @@
 
 import assert = require('assert');
 import events = require('events');
-import util = require ('util');
+import util = require('util');
 
 import blpapi = require('blpapi');
 import Promise = require('bluebird');
@@ -11,6 +11,7 @@ import debug = require('debug');
 
 // LOGGING
 var trace = debug('blpapi-wrapper:trace');
+var log = debug('blpapi-wrapper:debug');
 
 // TYPES
 export interface RequestCallback {
@@ -57,9 +58,8 @@ export class Session extends events.EventEmitter {
     // PRIVATE MANIPULATORS
     private listen(eventName: string, expectedId: number, handler: Function) {
         if (!(eventName in this.eventListeners)) {
-            trace(util.format('\'%s\' listener added', eventName));
+            trace('Listener added: ' + eventName);
             this.session.on(eventName, (function(eventName: string, m: any) {
-                trace(m);
                 var correlatorId = m.correlations[0].value;
                 this.eventListeners[eventName][correlatorId](m);
             }).bind(this, eventName));
@@ -72,7 +72,7 @@ export class Session extends events.EventEmitter {
     private unlisten(eventName: string, correlatorId: number) {
         delete this.eventListeners[eventName][correlatorId];
         if (isObjectEmpty(this.eventListeners[eventName])) {
-            trace(util.format('\'%s\' listener removed ', eventName));
+            trace('Listener removed: ' + eventName);
             this.session.removeAllListeners(eventName);
             delete this.eventListeners[eventName];
         }
@@ -81,6 +81,12 @@ export class Session extends events.EventEmitter {
     private requestHandler(cb: RequestCallback, requestId: number, m: any) {
         var eventType = m.eventType;
         var isFinal = (EVENT_TYPE.RESPONSE === eventType);
+
+        log(util.format('Response: %s|%d|%s',
+                        m.messageType,
+                        m.correlations[0].value,
+                        eventType));
+        trace(m);
 
         cb(null, m.data, isFinal);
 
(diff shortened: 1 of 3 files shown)


ericvw:
@ericlu88, @tmikov - when you get a chance, please provide your review & feedback.

tslib/blpapi-wrapper.ts
```diff
@@ -11,6 +11,7 @@ import debug = require('debug');
 
 // LOGGING
 var trace = debug('blpapi-wrapper:trace');
+var log = debug('blpapi-wrapper:debug');
 
 // TYPES
 export interface RequestCallback {
```
ericlu88, line 17: Maybe we should think of refactoring out the interfaces and util classes out to a separate file. I generally prefers one class per module. Just easier to track.

tslib/blpapi-wrapper.ts
```diff
@@ -44,21 +45,27 @@ function isObjectEmpty(obj: Object) {
...
-    private eventListeners: {[index: string]: any} = {};
+    private eventListeners: {[index: string]: {[index: number]: Function}} = {};
     private requests: {[index: string]: RequestCallback} = {};
     private services: {[index: string]: Promise<void>} = {};
     private correlatorId: number = 0;
-    private requestId: number = 0;
+    private stopped: Promise<void> = null;
 
     // PRIVATE MANIPULATORS
+    private nextCorrelatorId() {
```
ericlu88: To make it consistent, shall we add return type to all functions?
`private nextCorrelatorId() : void {`

tslib/blpapi-wrapper.ts
```diff
@@ -44,21 +45,27 @@ function isObjectEmpty(obj: Object) {
...
+    }
+
     private listen(eventName: string, expectedId: number, handler: Function) {
         if (!(eventName in this.eventListeners)) {
-            trace(util.format('\'%s\' listener added', eventName));
-            this.session.on(eventName, (function(eventName: string, m: any) {
-                trace(m);
+            trace('Listener added: ' + eventName);
+            this.session.on(eventName, ((eventName: string, m: any) => {
                 var correlatorId = m.correlations[0].value;
```
ericlu88, line 69: Maybe add some error handling for cases where the returned object does not have a correlations property or the returned correlation id does not corresponding to the one we sent. At least log it so it's easier to track down?

tslib/blpapi-wrapper.ts
ericvw: Yes, we should.  I will see if this can be enforced by `tslint` and make the change.

tslib/blpapi-wrapper.ts
ericvw, line 69: I will add an `assert` since this should never happen and would cause the following code to crash anyways.

tslib/blpapi-wrapper.ts
ericvw, line 17: I would like to hear @tmikov's feedback as well.  I have a couple of thoughts & considerations on this: First, we should keep in mind not only what should be going into files, but what also should be going together in modules.  TypeScript can do merge a module from multiple files, but this may be more confusing than having it co-located in the same file.  Secondly, I am a bit partial to keeping the definition of the interface close to where it is used; especially if it is used/required by a specific class - it seems reasonable to keep it with that class because the dependent code that uses the class will be nature need the interface as well (instead of importing it from somewhere else).  Third, util classes that are internal implementation details can stay within the module if it is small enough.  Open to more discussion this though.

ericlu88, line 17: For the third point, internal util classes can stay within the module, but I don't think they need to be exported.

tslib/blpapi-wrapper.ts
ericvw: `tslint` can catch these errors, however there seems to be a bug since it is complaining that a type is required for constructors, but TypeScript doesn't allow that.  I will address this specific issue for now and work with https://github.com/palantir/tslint to make this work properly with `constructor()` after more investigation.

ericvw:
Updated PR with changes addressed.  Awaiting further feedback/discussion on breaking out potentially the interfaces into separate files.

tslib/blpapi-wrapper.ts
ericvw: Turns out that PR palantir/tslint#245 addressed the bug already.  Once the next version of `tslint` is released we can bump-up to the latest version and enable these checks at lint time.

PR #8, opened 2015-01-08
Use " instead of ' for response property names.

yupinghu:
Make data uniform with the other properties in the response, which use ".


tslib/api-session.ts
```diff
@@ -91,7 +91,7 @@ export function makeHandler (): (req: OurRequest, res: OurResponse, next: Functi
                     if (needComma) {
                         chunk += ',';
                     }
-                    chunk += '\'data\':[';
+                    chunk += '\"data\":[';
```
ericvw: Since the outer string is encapsulated in single quotes, `'`, you can remove escape slash (i.e., `\`).

PR #12, opened 2015-01-13
Add configurations for service

ericlu88:
These commits address:
- Convert config.js to typescript so all of our sources are in ts. Rename `/tslib` directory to be `/lib`.
- Add configuration for the following field:
  - https:
    - enable: true/false
    - ca: path to the ca file
    - cert: path to the cert file
    - key: path to the key file
  - logging:
    - stdout: true/false for stdout
    - stdoutLevel: log level for stdout
    - logfile: path for the log file
    - logfileLevel: log level for file logging
  - service:
    - name: service name
    - version: service version
  - maxBodySize: Maximum size of the request body in byte
  - throttle:
    - burst: Throttle burst
    - rate: Throttle rate
- Change log property type in restify.d.ts, so that the JetBrain IDEs are not complaining.


yupinghu:
LGTM but evw should probably take a quick glance too since he's more familiar with the details.

ericvw:
LGTM

ericlu88:
Thanks guys!

Issue #13, opened 2015-01-13
HTTPS Client Certificates


ericvw:
This issue is to investigate and implement the usage of HTTPS client certificates as a means of authentication and identity management.

Some investigative questions to answer are:
- Do we have access to the underlying certificate information?
  - If so, what part of the certificate can we use to uniquely identify the user?
- How does revocation lists of certificates work with the http server?

Please feel free to add any other investigative questions, answers, and progress.

ericlu88:
Progress so far regarding getting access to underlying client cert info:

Getting access to the client cert in node is relatively easy. One just has to call `req.connection.getPeerCertificate()`, as noted in the [Node API doc](http://nodejs.org/api/http.html):

> With HTTPS support, use request.connection.verifyPeer() and request.connection.getPeerCertificate() > to obtain the client's authentication details.

The problem comes in when we try to do this in typescript. `req.connection` is type `net.Socket` when req is a **HTTP** request, and type `tls.ClearTextStream` when req is a **HTTPS** request. In order to get variant type notation in typescript, I upgrade the typescript compiler to 1.4.0 so that we could use **union** type(`req.connection: net.Socket|tls.ClearTextStream`). 

Within the extracted client certificate object, there's the **fingerprint** property that we could use to unique identify the user. I will start modifying the subscription code based on it.

tmikov:
You can always do a typecast in typescript.

I am more curious about how to invalidate a client certificate?

ericlu88:
Typecasting would only work in up/down casting, as explained in this [post](http://blogs.microsoft.co.il/gilf/2013/01/18/using-casting-in-typescript/). In this case, if the type annotation is `req.connection: net.Socket`, I get the following error when I try to cast it to `tls.ClearTextStream`:

> error TS2352: Neither type 'Socket' nor type 'ClearTextStream' is assignable to the other.
> Property 'authorized' is missing in type 'Socket'.

I will look into the client cert invalidation tomorrow.

tmikov:
I don't have time to read the link, but you can totally cast anything into anything.
var x : Cls1;
var y : Cls2;
y = <Cls2><any>x;

tmikov:
Stupid Github doesn't display < and > :-)

ericlu88:
I've tried the following example and it doesn't seem to work:

```
interface Dog {
    woo(): void;
}

interface Cat {
    meow(): void;
}

var d: Dog;
var c: Cat;

d.woo();    // OK
[...]

PR #21, opened 2015-01-14
Add logging for request body & client certificate

ericlu88:
This change addresses the following issues(relate to issue #13 ):
- Add a plugin class for custom restify middleware. Right now there's only a log middleware for logging request body and client cert details.
- Use the dev dependency typescript compiler in the Make file.
- Update typescript compiler to 1.4.0 so we could leverage the union type definition.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,12 @@ SRCS_TS += $(wildcard lib/*.ts)
 
 SRCS_JS_FROM_TS += $(patsubst %.ts,%.js,$(SRCS_TS))
 
-TSC = tsc
+TSC := $(shell npm bin)/tsc
 TSC_COMMON = --module commonjs --target ES5 --sourceMap --noImplicitAny
 TS_TO_JS = $(TSC) $(TSC_COMMON)
 TSC_TARGET = .tsc.d
 
-TSLINT = tslint
+TSLINT := $(shell npm bin)/tslint
 TSLINT_TARGET = .tslint.d
 
 RM ?= rm -f
--- a/index.ts
+++ b/index.ts
@@ -7,6 +7,7 @@ import bunyan = require('bunyan');
 import BAPI = require('./lib/blpapi-wrapper');
 import apiSession = require('./lib/api-session');
 import conf = require('./lib/config');
+import plugin = require('./lib/plugin');
 
 var logger: bunyan.Logger = bunyan.createLogger(conf.get('loggerOptions'));
 var session: BAPI.Session;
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ SRCS_TS += $(wildcard lib/*.ts)
 
 SRCS_JS_FROM_TS += $(patsubst %.ts,%.js,$(SRCS_TS))
 
-TSC = tsc
+TSC := $(shell npm bin)/tsc
```
ericvw: I think we will also need to do this for `TSLINT = tslint` line too.

index.ts
```diff
@@ -21,7 +22,7 @@ createSession()
     var server = restify.createServer(serverOptions);
 
     // Setup request logging
-    server.pre((req: restify.Request, res: restify.Response, next: restify.Next): any => {
+    server.pre(function pre(req: restify.Request, res: restify.Response, next: restify.Next): any {
```
ericvw, line 25: Why did this change away from a TypeScript anonymous function?

ericlu88, line 25: The restify trace log all middleware function names. The anonymous function will show up as ? in the log, so I decide to give it a name.

ericvw:
There are other changes going on that I don't think are meant to be part of this PR.  It seems like some of your union type changes got muddled in the commit.  Can you re-upload your branch with just the logging changes?

ericlu88:
The logging of client certificate depends on the union type definition, that's why they got muddled together.

ericvw:
Got it - will continue reviewing.

lib/plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+
+class Plugin {
+
+    // Custom logging plugin.
+    // Log client cert details, request body, etc
+    public static log(): (req: apiSession.OurRequest,
+                          res: apiSession.OurResponse,
+                          next: Function ) => void {
+
+        return function log( req: apiSession.OurRequest,
```
ericvw: Let's remove the space between the opening parens and the parameter. => `log(req:`

lib/config.ts
```diff
@@ -153,6 +165,16 @@ class Config {
                 header: res._headers
             };
         };
+        // Add client cert serializer
+        bunyan.stdSerializers['cert'] = function(cert) {
+            if (!cert) {
```
ericvw: This block was a bit confusing to read.  I would either have the second return in an `} else {` block or do the following:

``` typescript
return cert && {
    CN: cert.subject.CN,
    fingerprint: cert.fingerprint
}
```

lib/plugin.ts
ericlu88: done.

ericvw:
Going to do a test clone and install with this branch to ensure that the TypeScript installation works as expected.  Once everything checks out, we can squash all these commits together and then I will merge.

lib/config.ts
ericlu88: Done. Prefer the second approach.

ericvw:
lgtm - please squash and then I will merge.

ericlu88:
squash done.

PR #33, opened 2015-01-27
Add Subscription over long-polling and websocket

ericlu88:
This PR adds subscription support for blpapi-http via long-polling and websocket(both socket.io & ws implementation).


lib/blpapi-wrapper.ts
```diff
@@ -287,7 +287,7 @@ export class Session extends events.EventEmitter {
         return Promise.all(_.map(subscriptionsToServices(subscriptions), (uri): Promise<void> => {
             return this.openService(uri);
         })).then((): void => {
-            log('Subscribing to: ' + JSON.stringify(subscriptions));
+            //log('Subscribing to: ' + JSON.stringify(subscriptions));
```
ericlu88, line 290: I came across this interesting issue and welcome any input. Basically there's an error `Converting circular structure to JSON` thrown by calling `JSON.stringify(subscriptions)`

The subscriptions object looks like this:

```
[{ domain:
     { domain: null,
       _events: [Object],
       _maxListeners: 10,
       members: [Object] },
    _events: { data: [Function] },
    _maxListeners: 10,
    correlationId: 1,
    security: 'GOOG US Equity',
    fields: [ 'LAST_PRICE' ],
    options: undefined }]
```

The Subscription object is defined like this in typescript:

``` javascript
class Subscription extends events.EventEmitter  {
    correlationId: number;
    security: string;
    fields: string[];
    options: any;
    // constructor, etc...
}
```

So the circular reference comes from the nested `domain` properties. I did some research and here's what [Node](http://nodejs.org/api/domain.html#domain_class_domain) docs said:

screen shot 2015-01-26 at 4 39 37 pm
screen shot 2015-01-26 at 6 21 08 pm

Resitfy uses this domain functionality to handle errors, and because of how typescript translate class inheritance(it seems to me that it will always elevate base "class" property up to the derived class), `domain` becomes an own property of 'subscription`, thus making the`JSON.stringify` error.

Any thoughts?

ericvw, line 290: For this, we probably need to have the `blpapi-wrapper.ts` export `Subscription` as a class instead of an interface to enforce only the properties it expects.

ericvw:
Wow - this is a large change with a lot of different things going on.

I think we are going to have to break this down into smaller chunks.  First, let's start breaking out the type definitions updates and cleanup into a separate PR.

I need to spend some time digesting this some more and to think about how we can break this out into smaller PRs and/or commits.  @yupinghu, your help would be greatly appreciated :).

ericlu88:
Ok, I will create a separate PR for type definitions stuff.

ericvw:
@ericlu88, awesome - let me know if I can help in anyway with splitting out commits/changes.  While I am digesting, I will also start work on changing `interface Subscription` to `class Subscription`.

ericlu88:
@ericvw Can you elaborate a bit of your intended change of `class Subscription`? Note that I already created such class and implements `interface Subscription`.

Also, when you have time, could you take a look at the question about the **circular JSON error** I posted above? Appreciate your help.

ericlu88:
Create PR #34 for type definition changes.

ericvw:
@ericlu88 the `class Subscription` was in relation to the _circular JSON error_; however, the more I think about it.  The approach I was thinking about won't work since anyone could just extend the class and pass it through.  I need to think on this some more.

ericlu88:
Ideally the `domain, _events & _maxListeners` properties should be only living on `Subscription.prototype`(aka `EventEmitter`), not on the subscription object itself, right? The part I am confused is why typescript compiler elevate the properties to the parent object.

ericvw:
I did a force-push to `develop`, you may want to rebase this change against the latest `develop`.

ericlu88:
Rebased against latest develop. 

I will work on breaking up this PR into smaller chunk.

ericlu88:
Add PR #36 as the first chunk for this big change.

ericlu88:
Add PR #44 for the second chunk of this change.

ericlu88:
Close this big PR since it already get broken down.

PR #34, opened 2015-01-27
Update/Cleanup type declaration files

ericlu88:
This PR add/updates type declaration files needed by subscription changes.


typings/node/node.d.ts
```diff
@@ -291,7 +292,7 @@ declare module "http" {
         setEncoding(encoding?: string): void;
         pause(): void;
         resume(): void;
-        connection: net.Socket;
+        connection: net.Socket|tls.ClearTextStream;
```
ericlu88, line 295: This is a local change for now and is a pending PR for borisyankov/DefinitelyTyped#3517.

typings/optimist/optimist.d.ts
```diff
@@ -7,49 +7,49 @@
...
-        export interface Argv extends Object {
-            _: string[];
-        }
+		export interface Argv extends Object {
+			_: string[];
+		}
 
-        export var argv: any;
-    }
+		export var argv: any;
```
ericlu88, line 51: I couldn't find out any better way to typed the `argv` parameter. Welcome any suggestion.

typings/restify/restify.d.ts
```diff
@@ -194,6 +195,7 @@ declare module "restify" {
   export class ResourceNotFoundError { constructor(message: any); }
   export class WrongAcceptError { constructor(message: any); }
   export class UnsupportedMediaTypeError { constructor(message: any); }
+  export class RequestTimeoutError { constructor(message: any); }
```
ericlu88, line 198: Pending on borisyankov/DefinitelyTyped#3546

typings/optimist/optimist.d.ts
ericvw, line 51: I'll do some research as well on this.

ericvw:
I would suggest that we remove the type definitions that are needed for PR #33 since those are definitely required for that change.

This PR may be best characterized as cleanup and updates since the history will follow directly from the `develop` branch.

typings/socket.io/socket.io.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        emit(name: string, ...args: any[]): Socket;
+        join(name: string, fn?: Function): Socket;
+        leave(name: string, fn?: Function): Socket;
+        to(room: string): Socket;
+        in(room: string): Socket;
+
+        on(event: string, listener: Function): any;
+        broadcast: Socket;
+        volatile: Socket;
+        connected: boolean;
```
ericlu88: Pending on borisyankov/DefinitelyTyped#3547.

ericlu88:
Removed type definitions related to PR #33 and renamed this PR.

ericvw:
Thanks for trimming the PR down.

I think I have an elegant solution for type definition overrides and extensions.  Take a look at ericvw/blpapi-http@a841df7ca3a4913e2bb5c149082a045e0a381561 for extending other type definitions.

What I like about this type of approach is that we aren't customizing the type definitions we get from the [DefinitelyTyped](https://http://definitelytyped.org/) project via `tsd`.

What do you think?

ericvw:
I also noticed that the `tsd.d.ts` file and `tsd.json` are out of sync with respect to the ordering of the dependency manifest and the type definition references, respectively.  Did `tsd` rearrange the order of these files?  I thought `tsd` only appended new items.

In any event, it would be great if we could keep the ordering consistent across both files.  Since the `blpapi` definitions specific to this project, it probably makes sense to put that reference to the top of `tsd.d.ts` since calls to `tsd` appends to this file for newly installed definitions.

ericlu88:
Wow, this is cool. Definitely better than maintaining a local copy of all those type files.

I guess the only downside I see is when DefinitelyTyped add the same property eventually(by someone else) and we may end up with a type conflict error in compile time if we do a `tsd update`. Since we are not making customized typings files specific to our app, this could be a good work around while we submit PR and before it is accepted. After merged, we still should update via tsd and get rid of our local modification.

ericvw:
I did a force-push to `develop`, you may want to rebase this change against the latest `develop`.

ericlu88:
Type definition order cleaned-up.

ericvw:
Thanks for the ordering cleanup.  After rebasing against `develop`, I will merge this PR.

ericlu88:
Rebased.
@ericvw Do you want to create a separate PR for your changes ericvw/blpapi-http@a841df7ca3a4913e2bb5c149082a045e0a381561?

ericvw:
lgtm.  I will go ahead and create the PR for the override changes.

PR #35, opened 2015-01-27
restructure type definition files and directories

ericvw:
This change introduces a level of indirection for referencing type
definitions and a convention for extending/overriding definitions
provided by the [DefinitelyTyped](http://http://definitelytyped.org/)
project.

The primary motiviations for introducing another level of indirection
for referencing type definitions are:
1. Provide each `.ts` file stability for the `.d.ts` file it references.
2. The subdirectory `typings`, where the stable `.d.ts` reference is
   located, also serves as the directortory to place
   extensions/overrides to third-party type definitions.

The consequence is that each `.ts` files only needs to reference
`typings/definitions.d.ts`, which will remain stable in event of file
moves.

Additionally, overrides may be placed in the `typings` subdirectory and
need only to be referenced by `typings/definitions.d.ts`.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,4 +1,4 @@
-/// <reference path='typings/tsd.d.ts' />
+/// <reference path='typings/definitions.d.ts' />
 
 import restify = require('restify');
 import Promise = require('bluebird');
--- a/lib/typings/restify-override.d.ts
+++ b/lib/typings/restify-override.d.ts
@@ -1,0 +1,7 @@
+// Type definition overrides for restify
+
+declare module "restify" {
+    // borisyankov/DefinitelyTyped#3546
+    export class UnsupportedMediaTypeError { constructor(message: any); }
+}
+
--- a/typings/definitions.d.ts
+++ b/typings/definitions.d.ts
@@ -1,0 +1,1 @@
+/// <reference path='tsd.d.ts' />
--- a/lib/api-session.ts
+++ b/lib/api-session.ts
@@ -1,4 +1,4 @@
-/// <reference path='../typings/tsd.d.ts' />
+/// <reference path='typings/definitions.d.ts' />
 
 import assert = require('assert');
 import http = require('http');
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -1,4 +1,4 @@
-/// <reference path="../typings/tsd.d.ts" />
+/// <reference path='typings/definitions.d.ts' />
 'use strict';
 
 import assert = require('assert');
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,3 +1,5 @@
+/// <reference path='typings/definitions.d.ts' />
+
 import fs = require('fs');
 import path = require('path');
 import convict = require('convict');
--- a/lib/typings/blpapi.d.ts
+++ b/lib/typings/blpapi.d.ts
@@ -1,9 +1,4 @@
-// Type definitions for blpapi-node
-// Project: https://github.com/bloomberg/blpapi-node
-// Definitions by: Tzvetan Mikov <https://github.com/tmikov>
-// Definitions: https://github.com/borisyankov/DefinitelyTyped
-
-/// <reference path="../node/node.d.ts" />
+/// <reference path='definitions.d.ts' />
 
 declare module "blpapi" {
 
--- a/lib/typings/definitions.d.ts
+++ b/lib/typings/definitions.d.ts
@@ -1,0 +1,4 @@
+/// <reference path='../../typings/definitions.d.ts' />
+/// <reference path='blpapi.d.ts' />
+/// <reference path='optimist-override.d.ts' />
+/// <reference path='restify-override.d.ts' />
--- a/lib/typings/optimist-override.d.ts
+++ b/lib/typings/optimist-override.d.ts
@@ -1,0 +1,7 @@
+// Type definition overrides for optimist
+
+declare module "optimist" {
+    module optimist {
+        export var argv: any;
+    }
+}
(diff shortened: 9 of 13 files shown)


lib/typings/restify-override.d.ts
```diff
@@ -0,0 +1,6 @@
+// Type definition overrides for restify
+
+declare module "restify" {
+    export class UnsupportedMediaTypeError { constructor(message: any); }
```
ericvw, line 5: Should I add a comment reference to borisyankov/DefinitelyTyped#3546?

index.ts
```diff
@@ -1,4 +1,4 @@
-/// <reference path='typings/tsd.d.ts' />
+/// <reference path='typings/definitions.d.ts' />
```
ericlu88, line 1: Since `index.ts` is outside lib folder, should it be `lib/typings/definitions.d.ts`?

typings/definitions.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path='tsd.d.ts' />
```
ericlu88, line 1: What's the need of this typing file? My understanding of this structure is that the `typings` directory contains all type definition get from DefinitelyTyped through `tsd`, and the `lib/typings` subdirectory contains our overrides and extenstions, so the `lib/typings/definitions.d.ts` only needs to reference 'typings/tsd.d.ts', right?

index.ts
ericvw, line 1: The intent is for each `.ts` file to only reference `typings/definitions.d.ts` for stability.  Having it depend upon the path prefix `lib/` would require the reference to change if `index.ts` moved.

ericlu88:
Cool. We probably should also add a `README` file or something within the `lib/typings` subdirectory to mandate all typing overrides that has a pending PR has to be removed once the PR is merged and update the type definition in `typing` directory via `tsd`.

index.ts
ericlu88, line 1: Ok, but the side effect would be index.ts is looking for typing in `typings/definition.d.ts`, which only has the DefinitelyTyped stuff, not our overrides?

typings/definitions.d.ts
ericvw, line 1: Similar reasoning to my comment above - if `lib/` had more hierarchy, each directory would have a `typings` directory with `definitions.d.ts` that would contain at least:

``` ts
/// <reference path='../../typings/defintions.d.ts' />
```

To refer to the parent directory's typings.

index.ts
ericvw, line 1: It would get our overrides because it `import`'s our local modules, which has the reference to the type definitions :).  At least that is my understanding of what `tsc` is doing.

ericvw:
Let me know if you prefer a `README` of some sort for this PR and I can write one up.  Otherwise, I can provide one in a separate PR.  I am not too overly concerned because these overrides should be short-lived or changes can away upstream changes to land first.

Let me know either way.

index.ts
ericlu88, line 1: Well, take an example if we are gonna use `restify.UnsupportedMediaTypeError` in `index.ts`. `index.ts` has an `import restify = require('restify')`, and it only takes typing from `tsd`, I would guess `tsc` will say can't find. Can you give it a shot in `index.ts`, like in line **101**, substitue `InternalError` with `UnsupportedMediaTypeError`, would it compile?

ericlu88:
I lean toward a simple `README` and maybe reference this PR in that file. Thanks!

index.ts
ericvw, line 1: Ahh!  I get what you are saying now.

It does compile right now only because it is getting the type definitions from another import :).

I guess the question is whether we should scope overrides closest to usage or the third-party type definitions.  What do you think?

ericvw:
I'll work on some form of `README` document.  Let's hold off merging until one is provided in this PR, but let's continue other discussions.

index.ts
ericlu88, line 1: Well, I don't really understand the two choices you mentioned, but my preference would be a centralize type definition file that blends the `tsd` types and our overrides, and every `ts` file reference the same file. I am not too strong about the `path` stability issue, since refactoring the path should be part of directory restructuring?

ericlu88:
BTW, it just occurs to me that whether this approach can do **override**. [Stackoverflow](http://stackoverflow.com/questions/23217334/how-do-i-extend-a-typescript-class-definition-in-a-separate-definition-file) seems to suggest this approach only works if you try to extend the typings, but will throw `Duplicate Identifier` error if you try to override it. Can you try the `node` definition override we intend to add and see if it compiles?

ericlu88:
I tried to add the following `node-override.d.ts` file but get a `error TS2300: Duplicate identifier 'connection'`:

``` typescript
// Type definition overrides for node

declare module "http" {
    import events = require("events");
    import net = require("net");
    import stream = require("stream");
    import tls = require("tls");

    // borisyankov/DefinitelyTyped#3517
    export interface ServerRequest extends events.EventEmitter, stream.Readable {
        connection: net.Socket|tls.ClearTextStream;
    }
}
```

ericvw:
I get the same error too.  This needs a bit more thought.

What I was hoping to accomplish was the following:
1. Keep `tsd.d.ts` completely managed by `tsd`.
2. Provide some sanity to the paths used for `/// <reference path='../../../../../ />` so they that '.ts' files wouldn't need to be touched if they got moved.
3. Provide a mechanism to **extend** type definitions (i.e., for `optimist`)
4. Provide a mechanism for overrides, which we found out doesn't work.

Seeing how **4** cannot be accomplished, I am leaning towards either awaiting upstream definitions being patched and integrated back via `tsd` or we dupe the definition with our customizations and constantly update `tsd.json` to note whether we are using `tsd` for definitions (i.e., removing the definition/commit in `tsd.json` so we don't accidentally overwrite our customizations).

ericlu88:
The other alternative would be maintaining a folk everytime we need to extend/override some typing files, and let `tsd.json` file to reference our folk rather than the master branch. This approach also needs us to constantly update `tsd.json` once the PR is merged.

PR #36, opened 2015-01-27
Refactor blpapi-http

ericlu88:
This PR refactors code on top of develop branch:
- Refactor the route handling code(`onRequest()`) out to a separate class `requestHandler.ts`
- Refactor `api-session.ts` into typescript class. It will hold a static instance of `blpSession`, and attach it to every incoming request in `handleRequest()` middleware. 
- Add `interface.ts` to capture all our interfaces.
- Move request logging into `plugin.log()` function, which will be called at a later stage than `server.pre()`. This action will results in logging the request id with the request details.
- Move retrieving & logging client certificate into two steps, where retrieving is happening in `apiSession.handleRequest()` and logging is in `plugin.log()`. Use typescript 1.4 union type notation and syntax.
- Add `sessionOptions` into `conf` class for consistency.
- Remove unnecessary boolean flag `sessConnected` in `index.ts`.
- Rename `session` to `blpSession` in `index.ts`.
- Change `testRequest.js` to use `https` instead of `http`(and some formatting issue).

This PR serves as the starting block for the much larger changes in #33 while it does not add any functionality related to subscription.


lib/RequestHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+import apiSession = require('./api-session');
+import BAPI = require('./blpapi-wrapper');
+
+export = RequestHandler;
+
+class RequestHandler {
+    private blpSession: BAPI.Session;
+
+    constructor(blpsess: BAPI.Session) {
+        this.blpSession = blpsess;
```
ericvw: What happens when `blpsess` emits a `SessionTerminated` event?  It seems there is an attempt [here](https://github.com/ericlu88/blpapi-http/commit/4c449c45dce11e653116b1fef5634b048d12b224#diff-ed009b6b86b017532ef0489c77de5100R79).  However, that reassigns the global variable in `index.ts` without resetting the `blpSession` member variable within an instance of a `RequestHandler`.

ericvw: I have been reading up on [res.locals](http://expressjs.com/api.html#res.locals) from the [Express](http://expressjs.com/) project and it would be great to have something like this to have some middleware to tack-on `blpapi.Session` for further handlers to use.

ericlu88: I moved the instantiation of `requesthandler` inside `createSession` so it would hold a substituted instance of `blpSession` in case `sessionTerminated`. Good catch, thanks!

I don't fully understand the use of `res.locals`. Are you suggesting tacking the `blpSession` onto the `req/res`? If so, I don't particular see any benefit of it since we are only use **1** session object. Is there any performance implication for using closure rather than direct property lookup? The only reason I am not using early binding is that `bind` will create an anonymous function, which would result in a `?` in restify's trace log.

yupinghu:
I think this latest commit fixes the session terminated issue (and generally looks ok to me); evw, is this what you were thinking was the problem?

ericvw:
lgtm.  Would appreciate @yupinghu's feedback as well.  Once that let's good, let's squash some commits down and amp-up the commit message to describe the high-level details.

lib/middleware/blp-session.ts
```diff
@@ -0,0 +1,59 @@
...
+            });
+
+        SESSION_STORE[sessionId] = p;
+    }
+
+    return SESSION_STORE[sessionId];
+}
+
+export function getSession(req: interfaces.IOurRequest,
+                           res: interfaces.IOurResponse,
```
yupinghu: This param is unused, any reason to retain it here?

yupinghu:
LGTM other than the ' vs " thing we discussed. :)

ericlu88:
Squashed & change all reference tag to use single quote.

PR #37, opened 2015-01-30
expose `Subscription` as a class instead of an interface

ericvw:
When `interface Subscription` was exposed, callers could add their own
properties to the subscription object that would be passed-into the
blpapi wrapper.  This caused problems with the internal logging for
calls to `subscribe()` and `unsubscribe()` since additional properites
were being serialized.  To prevent this from happening, the wrapper now
exposes `Subscription` as a **class** and provides a `private toJSON()`
method for `JSON.stringify` to only serialize the properties that the
wrapper cares about for logging.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -19,10 +19,33 @@ export interface RequestCallback {
     (err: Error, data?: any, isFinal?: boolean): void;
 }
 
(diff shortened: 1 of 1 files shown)


yupinghu:
LGTM

ericlu88:
LGTM

PR #38, opened 2015-02-02
Add subscription via websocket(socket.io & ws)

ericlu88:
This PR add subscription functionality to blpapi-http via websocket(`socket.io` & `ws`).

Some implementation notes:
- Create a common interface `ISocket` and two wrapper classes `socket-io-wrapper.ts` & `ws-wrapper.ts`. This unify the interfaces exposed by two libraries and enable us to create a single business handling code `websocket-handler.ts`.
- `subscription-sotre.ts` use generics to enable child class supply data item types.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -7,13 +7,16 @@ import conf = require('./lib/config');
 import blpSession = require('./lib/middleware/blp-session');
 import requestHandler = require('./lib/middleware/request-handler');
 import util = require('./lib/middleware/util');
+import sio = require('socket.io');
+import webSocket = require('ws');
+import webSocketHandler = require('./lib/websocket/websocket-handler');
 
 var logger: bunyan.Logger = bunyan.createLogger(conf.get('loggerOptions'));
 
 // Create server.
 var serverOptions = conf.get('serverOptions');
 serverOptions.log = logger;     // Setup bunyan logger
-var server = restify.createServer(serverOptions);
+var server: restify.Server = restify.createServer(serverOptions);
 
 // Middleware
 server.pre(util.resetContentType);
@@ -43,3 +46,25 @@ server.on('listening', (): void => {
     logger.info('http server listening on', conf.get('port') );
 });
 server.on('after', util.after);
+
+// Socket.IO
+if (conf.get('websocket.socket-io.enable')) {
+    var serverSio: restify.Server = restify.createServer(conf.get('serverOptions'));
+    serverSio.listen(conf.get('websocket.socket-io.port'));
+    serverSio.on('listening', (): void => {
+        logger.info('socket.io server listening on', conf.get('websocket.socket-io.port') );
+    });
+    var io: SocketIO.Namespace = sio(serverSio.server).of('/subscription');
+    io.on('connection', webSocketHandler.sioOnConnect);
+}
+
+// ws
+if (conf.get('websocket.ws.enable')) {
+    var serverWS: restify.Server = restify.createServer(conf.get('serverOptions'));
+    serverWS.listen(conf.get('websocket.ws.port'));
+    serverWS.on('listening', (): void => {
+        logger.info('ws server listening on', conf.get('websocket.ws.port') );
+    });
+    var wss = new webSocket.Server({server: serverWS.server});
+    wss.on('connection', webSocketHandler.wsOnConnect);
+}
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -48,7 +48,7 @@ export function get(name: string): any
             env: 'BLPAPI_HTTP_PORT',
             arg: 'port'
         },
-        'expiration' : {
+        'expiration': {
             doc: 'Auto-expiration period of blpSession in seconds',
             format: 'integer',
             default: 5
@@ -149,6 +149,36 @@ export function get(name: string): any
                 format: 'integer',
                 default: 50,
                 arg: 'throttle-rate'
+            }
+        },
+        'websocket': {
+            'socket-io': {
+                'enable': {
+                    doc: 'Boolean option to control whether to run socket.io server',
+                    format: Boolean,
+                    default: true,
+                    arg: 'websocket-socket-io-enable'
+                },
+                'port': {
+                    doc: 'The socket io port to listen on',
+                    format: 'port',
+                    default: 3001,
+                    arg: 'websocket-socket-io-port'
+                },
+            },
+            'ws': {
+                'enable': {
+                    doc: 'Boolean option to control whether to run ws server',
+                    format: Boolean,
+                    default: true,
+                    arg: 'websocket-ws-enable'
+                },
+                'port': {
+                    doc: 'The ws port to listen on',
+                    format: 'port',
+                    default: 3002,
+                    arg: 'websocket-ws-port'
+                },
             }
         }
     });
--- a/lib/interface.ts
+++ b/lib/interface.ts
@@ -1,7 +1,7 @@
 /// <reference path='../typings/tsd.d.ts' />
 
-import Promise = require('bluebird');
 import restify = require('restify');
+import bunyan = require('bunyan');
 import BAPI = require('./blpapi-wrapper');
 
 export interface IOurRequest extends restify.Request {
@@ -13,4 +13,14 @@ export interface IOurResponse extends restify.Response {
     sendChunk?: (data: any) => Promise<void>;
     sendEnd?: (status: any, message: string) => Promise<void>;
     sendError?: (err: any, where: string, reason?: any) => Promise<void>;
+}
+
+export interface ISocket extends NodeJS.EventEmitter {
+    log: bunyan.Logger; // Add bunyan logger to socket
+    blpSession: BAPI.Session;  // Add blpSession to socket
+    isConnected(): boolean;
+    getIP(): string;
+    getCert(): any;
+    disconnect(): void;
+    send(name: string, ...args: any[]): void;
 }
--- a/lib/middleware/blp-session.ts
+++ b/lib/middleware/blp-session.ts
@@ -1,5 +1,6 @@
 /// <reference path='../../typings/tsd.d.ts' />
 
+import Promise = require('bluebird');
 import bunyan = require('bunyan');
 import restify = require('restify');
 import blpapi = require('../blpapi-wrapper');
@@ -8,10 +9,7 @@ import interfaces = require('../interface');
 
 // CONSTANTS
 var DEFAULT_ID = '__default__';
-var DEFAULT_SESSION_OPTIONS = {
-    serverHost: conf.get('api.host'),
-    serverPort: conf.get('api.port')
-};
+var DEFAULT_SESSION_OPTIONS = conf.get('sessionOptions');
 
 // GLOBALS
 var SESSION_STORE: {[index: string]: Promise<blpapi.Session>} = {};
@@ -55,5 +53,18 @@ export function getSession(req: interfaces.IOurRequest,
         }).catch((err: Error): void => {
             req.log.error(err, 'Error getting blpSession.');
             return next(new restify.InternalError(err.message));
+        });
+}
+
+export function getSocketSession(socket: interfaces.ISocket): Promise<interfaces.ISocket>
+{
+    return createSession(DEFAULT_ID)
+        .then((session: blpapi.Session): interfaces.ISocket => {
+            socket.log.debug('blpSession retrieved.');
+            socket.blpSession = session;
+            return socket;
+        }).catch((err: Error): any => {
+            socket.log.error(err, 'Error getting blpSession.');
+            throw err;
         });
 }
--- a/lib/subscription/subscription-store.ts
+++ b/lib/subscription/subscription-store.ts
@@ -1,0 +1,72 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import _ = require('lodash');
+import Subscription = require('./subscription');
+
+export = SubscriptionStore;
+
+class SubscriptionStore<T extends Subscription> {
+
+    protected subscriptions: {[key: number]: T} = {};
+
+    get size(): number {
+        return _.keys(this.subscriptions).length;
+    }
+
+    get(cid: number): T {
+        return this.has(cid) ? this.subscriptions[cid] : undefined;
+    }
+
+    getAll(): T[] {
+        return _.values(this.subscriptions);
+    }
+
+    add(subscription: T): boolean {
+        if (this.has(subscription.correlationId)) {
+            return false;
+        }
+        this.subscriptions[subscription.correlationId] = subscription;
+        return true;
+    }
+
+    has(value: number): boolean;
+    has(value: T): boolean;
+    has(value: any): boolean {
+        if (typeof value === 'number') {
+            return _.has(this.subscriptions, value);
+        }
+        else if (typeof value === 'object') {
+            return _.has(this.subscriptions, value.correlationId);
+        }
+        return false;
+    }
+
+    delete(value: number): boolean;
+    delete(value: T): boolean;
+    delete(value: any): boolean {
+        if (typeof value === 'number') {
+            if (!this.has(value)) {
+                return false;
+            }
+            delete this.subscriptions[value];
+            return true;
+        }
+        else if (typeof value === 'object') {
+            if (!this.has(value)) {
+                return false;
+            }
+            delete this.subscriptions[value.correlationId];
+            return true;
+        }
+        return false;
+    }
+
+    clear(): void {
+        _.forOwn(this.subscriptions, (sub: T,
+                                      key: string): void => {
+            sub.removeAllListeners();
+        });
+        this.subscriptions = {};
+    }
+
+}
--- a/lib/subscription/subscription.ts
+++ b/lib/subscription/subscription.ts
@@ -1,0 +1,19 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import BAPI = require('../blpapi-wrapper');
+
+export = Subscription;
+
+class Subscription extends BAPI.Subscription {
+
+    correlationId: number;
+
+    constructor(cid: number,
+                security: string,
+                fields: string[],
+                options?: any) {
+        super(security, fields, options);
+
+        this.correlationId = cid;
+    }
+}
--- a/lib/websocket/websocket-handler.ts
+++ b/lib/websocket/websocket-handler.ts
@@ -1,0 +1,250 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import _ = require('lodash');
+import Promise = require('bluebird');
+import bunyan = require('bunyan');
+import uuid = require('node-uuid');
+import SocketIO = require('socket.io');
+import webSocket = require('ws');
+import Subscription = require('../subscription/subscription');
+import SubscriptionStore = require('../subscription/subscription-store');
+import conf = require('../config');
+import Interface = require('../interface');
+import blpSession = require('../middleware/blp-session');
+import SocketIOWrapper = require('./socket-io-wrapper');
+import WSWrapper = require('./ws-wrapper');
+
+// GLOBAL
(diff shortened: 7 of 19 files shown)


lib/middleware/blp-session.ts
```diff
@@ -57,3 +54,48 @@ export function getSession(req: interfaces.IOurRequest,
...
+            socket.blpSession = session;
+            return Promise.resolve(socket);
+        }).catch((err: Error): any => {
+            socket.log.error(err, 'Error getting blpSession.');
+            throw err;
+        });
+}
+
+/*
+export function getSession(req: interfaces.ISocket) : Promise<interfaces.ISocket>;
```
ericlu88: Not sure which approach is better so I leave the function overloading implementation below. I personally think it is ugly and prefer the separate function way.

lib/config.ts
```diff
@@ -150,6 +150,36 @@ export function get(name: string): any
                 default: 50,
                 arg: 'throttle-rate'
             }
+        },
+        'webSocket': {
```
ericvw: For options, especially for command-line arguments, let's go with `websocket`.  I know `web` and `socket` are distinct words, but this is also for configuration and aren't code variables.

lib/middleware/blp-session.ts
```diff
@@ -21,7 +18,7 @@ var LOGGER: bunyan.Logger = bunyan.createLogger(conf.get('loggerOptions'));
 function createSession(sessionId: string): Promise<blpapi.Session>
 {
     if (!(sessionId in SESSION_STORE)) {
-        var s = new blpapi.Session(DEFAULT_SESSION_OPTIONS);
+        var s = new blpapi.Session(conf.get('sessionOptions'));
```
ericvw: Since the `sessionOptions` configuration doesn't change at runtime, let's cache this as an initialization constant under the `// CONSTANTS` section.

lib/config.ts
ericlu88: Fixed.

lib/middleware/blp-session.ts
ericlu88: Well, it is already cached inside `conf` module. Is it necessary to create another copy here?

lib/strMap.ts
```diff
@@ -0,0 +1,70 @@
+/// <reference path='../typings/tsd.d.ts' />
+
+export = StrMap;
```
ericvw: What is the motivation for `StrMap`?  I would just use an `Object` as a dictionary and save having an alternative implementation.  ES6 maps are on their way soon hopefully with the next release of NodeJS.

index.ts
```diff
@@ -43,3 +46,25 @@ server.on('listening', (): void => {
...
+        logger.info('socket.io server listening on', conf.get('webSocket.socket-io.port') );
+    });
+    var io: SocketIO.Namespace = sio(serverSio.server).of('/subscription');
+    io.on('connection', webSocketHandler.sioOnConnect);
+}
+
+// ws
+if (conf.get('websocket.ws.enable')) {
+    var serverWS: any = restify.createServer(conf.get('serverOptions'));
+    serverWS.listen(conf.get('webSocket.ws.port'));
```
ericvw: We can have this discussion later, but it may be convenient to have the the `ws`, standard WebSocket protocol, share the same port as the http server (or have it be configurable that way).

lib/strMap.ts
ericlu88: It's a util class that Tzvetan wrote and I found it quite useful. I did not know a ES6 map is coming, and would agree with you to not having a separate implementation. I will change it to a raw `Object`.

index.ts
```diff
@@ -43,3 +46,25 @@ server.on('listening', (): void => {
     logger.info('http server listening on', conf.get('port') );
 });
 server.on('after', util.after);
+
+// Socket.IO
+if (conf.get('websocket.socket-io.enable')) {
+    var serverSio: any = restify.createServer(conf.get('serverOptions'));
```
ericvw: Is there a valid type definition for `serverSio`?

ericvw: Otherwise, does TypeScript infer the type in absence of this (preferred).

lib/middleware/blp-session.ts
ericvw: Looking at the code from a glance, one may think that this could change for each request that comes through unless they dig deeper and look at `lib/config.ts`.  There is nothing logically wrong with this line of code; it may be confusing on what it conveys without some global knowledge of the rest of the code base.

index.ts
```diff
@@ -43,3 +46,25 @@ server.on('listening', (): void => {
...
+    serverSio.on('listening', (): void => {
+        logger.info('socket.io server listening on', conf.get('websocket.socket-io.port') );
+    });
+    var io: SocketIO.Namespace = sio(serverSio.server).of('/subscription');
+    io.on('connection', webSocketHandler.sioOnConnect);
+}
+
+// ws
+if (conf.get('websocket.ws.enable')) {
+    var serverWS: any = restify.createServer(conf.get('serverOptions'));
```
ericvw: I would forgo the type definition if possible unless `tsc` is complaining about this.

lib/config.ts
```diff
@@ -150,6 +150,36 @@ export function get(name: string): any
...
                 arg: 'throttle-rate'
             }
+        },
+        'websocket': {
+            'socket-io': {
+                'enable': {
+                    doc: 'Boolean option to control whether to run socket.io server',
+                    format: Boolean,
+                    default: true,
+                    arg: 'webSocket-socket-io-enable'
```
ericvw: My apologies for not being clear in my comment earlier.  These 4 `arg:` property values can start with `websocket-*` as well since they are CLI arguments.

ericlu88:
Some outstanding problem/questions that I noticed while testing:
- When a socket is connected and I kill the underlying `blpSession`, I did not received a `SessionTerminated` events. The event is received if there's no socket connected. 
- Since we are using one `blpSession`, whenever a new client subscribe to a security, a `summary` data will be sent to all clients that are subscribing to the same security.

lib/strMap.ts
ericlu88: Implemented as suggested.

ericvw:
@ericlu88 Add GitHub issues so we can keep track of pinning down the `SessionTerminated`/socket issue.  I don't think it is that critical enough to hold up merging this PR.

For the `summary` data, I think this may be OK.  Once we have authorization and a way to track identity, these would be separate for each user and I would expect the Bloomberg Open API to broadcast the `summary` data only to the same authorized identity.

lib/middleware/blp-session.ts
ericvw: Definitely let's go with the separate function.

lib/subscription/subscription-store.ts
```diff
@@ -0,0 +1,72 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import _ = require('lodash');
+import Subscription = require('./subscription');
+
+export = SubscriptionStore;
+
+class SubscriptionStore<T extends Subscription> {
```
ericvw, line 8: This is similar in functionality to `Object`, I would almost argue that we could forgo this class as well.  Thoughts?

ericvw, line 8: If we do decide to keep this, I would make this a normal class and remove the generic since it **only** operations on `Subscription`.

lib/middleware/blp-session.ts
ericlu88: Done.

lib/subscription/subscription.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import BAPI = require('../blpapi-wrapper');
+
+export = Subscription;
+
+class Subscription extends BAPI.Subscription {
+
+    correlationId: number;
+
+    constructor( cid: number,
```
ericvw: Remove the space: `constructor( cid: number,` => `constructor(cid: number,` and indent remaining arguments accordingly.

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+                socket.emit('err', { message: 'Unexpected error: ' + err.message });
+                socket.disconnect();
+            }
+        });
+}
+
+function initialize(socket: Interface.ISocket): Promise<Interface.ISocket>
+{
+    // Setup logging for socket
+    socket.log = LOGGER.child({req_id: uuid.v4()});
```
ericvw: To make the `ISocket` abstraction cleaner, I would recommend that callers call a function to setup logging instead of directly modifying/setting the `log` property.

lib/interface.ts
```diff
@@ -14,3 +14,14 @@ export interface IOurResponse extends restify.Response {
...
+
+export interface ISocket {
+    log: bunyan.Logger; // Add bunyan logger to socket
+    blpSession: BAPI.Session;  // Add blpSession to socket
+    isConnected(): boolean;
+    getIP(): string;
+    getCert(): any;
+    emit(name: string, ...args: any[]): void;
+    disconnect(): void;
+    on(event: string, listener: Function): void;
```
ericvw: I think this interface can just extend `EventEmitter` from the node `events` module.

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+    // Subscribe
+    socket.on('subscribe', (data: Object[]): void => {
+        socket.log.info('Subscribe request received');
+        // Log body if configured
+        if (conf.get('logging.reqBody')) {
+            socket.log.debug({body: data}, 'Subscribe body.');
+        }
+
+        // Validate input options
+        var subscriptions : Subscription[] = [];
```
ericvw: => `var subscriptions:`

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+        if (!data.length) {
+            socket.log.debug('No valid subscriptions found.');
+            socket.emit('err', { message: 'No valid subscriptions found.' });
+            return;
+        }
+        data.forEach((s: any): void => {
```
ericvw: Is this `Subscription`?

ericvw: Ahh - this is the JSON object passed from the client.  So `Object`?

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+                return;
+            }
+            if (receivedSubscriptions.has(s.correlationId)) {
+                socket.log.debug('Correlation Id already exists.');
+                socket.emit('err', { message: 'Correlation Id already exists.' });
+                return;
+            }
+
+            var sub = new Subscription(s.correlationId,
+                                            s.security,
```
ericvw: Re-indent arguments.

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+                // TODO: Acknowledgement callback function?
+                socket.emit('data', {'correlationId' : sub.correlationId,
+                                     'data' : data});
+                socket.log.info('Data sent');
+            });
+        });
+
+        // Subscribe user request through blpapi-wrapper
+        socket.blpSession.subscribe(subscriptions)
+        .then((): void => {
```
ericvw: Indent in by 4 spaces :)

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,238 @@
...
+            // Otherwise, validate all correlation Id
+            // Will unsubscribe only the valid correlation Id, ignore the rest
+            data.correlationIds.forEach((cid: number): void => {
+                if (activeSubscriptions.has(cid)) {
+                    subscriptions.push(activeSubscriptions.get(cid));
+                }
+            });
+        }
+
+        if (!subscriptions.length) {
```
ericvw: I think this makes sense only in the context if correlation ids were specified.  I would suggest moving this into the `else` block and create an `else` block for the `if (activeSubscriptions.has(cid)) {` to generate a string of the invalid correlation ids found` for error reporting.

ericvw: I would appreciate @yupinghu's thoughts on whether an empty list of subscriptions is an error or not.  I am leaning towards that it is a no-op.

lib/middleware/blp-session.ts
ericlu88: Got ya

lib/websocket/websocket-handler.ts
ericvw: I'm going to see if `tslint` can capture this.

index.ts
ericlu88: I need to expose one more property of the restify typing file in order to make it work. I will submit a PR for this change later.

lib/config.ts
ericlu88: Fixed.

ericlu88:
Created issue #39 to track the `SessionTerminated` problem

lib/subscription/subscription-store.ts
ericlu88, line 8: This class encapsulate some common data retrieval/look up functions and I would suggest to keep it. It will make more sense for the long-polling version where I extend from this class and add buffering related operations into it. The generic is needed so the child class can supply the type of `Subscription` they operate on. This is what I have for the long-polling:

``` typescript
class Subscription {}
class SubscriptionWithBuffer extends Subscription{}

class SubscriptionStore<T extends Subscription> {}
class SubscriptionStoreWithBuffer extends SubscriptionStore<SubscriptionWithBuffer>
```

lib/websocket/websocket-handler.ts
ericlu88: @ericvw Can you elaborate a bit on this? How about making the logger a part of the wrapper constructor?

lib/subscription/subscription-store.ts
ericvw, line 8: I'll defer on this until the polling changes come in.  For the rest of the file, let's clean up she spacing around the parameters in the declarations and the the type declarations (i.e., remove the space padding).  I will have PR soon so that `tslint` captures these things.

lib/websocket/websocket-handler.ts
ericvw: I am all for a constructor or method.  If it is a one-time initialization that cannot change once a class that implements `ISocket` is constructed, then definitely in the the constructor.  If it is expected that this could change after construction, then a method.

lib/websocket/websocket-handler.ts
ericvw: It can and I will have a PR on this soon.

lib/interface.ts
ericlu88: Well, doing so will force the wrapper implement all the `EventEmitter` functions, which I am not sure if it is worth it. I only use `on` & `emit` though.

ericvw: You just also extend from `EventEmitter` in the class declaration.  You shouldn't need to reimplement all the methods.

lib/websocket/websocket-handler.ts
ericlu88: Changed to let `constructor` take `logger` as a one-time initialization, and expose `log` as only a getter.

lib/subscription/subscription-store.ts
ericlu88, line 8: Sorry,  I don't know where they came in. Fixed.

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,236 @@
...
+        socket.log.info('Subscribe request received');
+        // Log body if configured
+        if (conf.get('logging.reqBody')) {
+            socket.log.debug({body: data}, 'Subscribe body.');
+        }
+
+        // Validate input options
+        var subscriptions : Subscription[] = [];
+        if (!data.length) {
+            socket.log.debug('No valid subscriptions found.');
```
ericvw, line 84: PR #40 adds support for additional lint checks to catch this going forward.

lib/websocket/websocket-handler.ts
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        var subscriptions : Subscription[] = [];
+
+        // If no correlation Id specified
+        // the default behavior is to unsubscribe all active subscriptions
+        if (!data
+            || !_.has(data, 'correlationIds')
+            || !_.isArray(data.correlationIds)
+            || !data.correlationIds.length) {
```
ericvw: Bring `{` on the next line due to conventions discussed previously.

lib/interface.ts
ericlu88: Normally I would expect this from a `EventEmitter`:

``` typescript
class Foo extends events.EventEmitter{}
var foo = new Foo();
foo.on('bla', function listener(){});
foo.emit('bla');
```

I would expect `listener` to be called. But this is not the case for our socket wrapper.

lib/middleware/blp-session.ts
```diff
@@ -57,3 +55,16 @@ export function getSession(req: interfaces.IOurRequest,
...
         });
 }
+
+export function getSocketSession(socket: interfaces.ISocket) : Promise<interfaces.ISocket>
+{
+    return createSession(DEFAULT_ID)
+        .then((session: blpapi.Session): Promise<interfaces.ISocket> => {
+            socket.log.debug('blpSession retrieved.');
+            socket.blpSession = session;
+            return Promise.resolve(socket);
```
ericvw: This can just be `return socket` I believe because `.then()` returns a promise.

lib/websocket/websocket-handler.ts
yupinghu: As we discussed in chat, I could go either way here. Maybe we should have some debug logging message for now, not have it be an error, and see how often this comes up in practice?

lib/websocket/websocket-handler.ts
ericlu88: I should probably just remove this part of the code since it is attaching stuff to the actual data object sent back.

lib/websocket/websocket-handler.ts
ericlu88: I am a bit confused about what you guys suggesting, but this is the rule that I implemented:
- If no correlation ids specified(this includes no `correlationIds` property in the JSON object, or the `correlationIds` property is not an `Array`, or an empty list of `correlationIds`)
  - Unsubscribe all active subscriptions
- If correlation ids specified, validate all received correlation ids
  - If found any valid id, ignore the invalid ones and only unsubscribe the valids
  - If all correlation ids received are invalid, send back error

Let me know if this is good or we want something else.

lib/websocket/websocket-handler.ts
ericlu88: I move the logical operation to the end of line. However, tslint doesn't seem to like `{` onto the next line?

index.ts
ericlu88: Add PR borisyankov/DefinitelyTyped#3594 for this change.

lib/websocket/websocket-handler.ts
ericvw: The only difference @yupinghu and I were thinking was that if the list is empty, then no-op instead of unsubscribe all.

ericvw: My proposal:
- The JSON object must be well-formed (has expected properties), otherwise error
- If the list of `correlationIds` are empty, either unsubscribe all or no-op (we should decide which one).
- In the event there are `correlationIds`
  - If valid, continue
  - If invalid, whether one up to all, ignore

ericlu88: Implemented as the following(slightly different from @ericvw suggestion):
- The JSON object must be well-formed (has expected properties), otherwise error
- If the list of `corrleationIds` are empty, error(consistent with subscribe)
- In the event there are `correlationIds`
  - If any invalid, error
  - If all valid, continue

ericvw: :+1:

ericvw:
lgtm - please squash down into a single commit.  Also, for the single commit, please add more details in the commit body of the message that highlights the top-level details (e.g., ISocket to homogenize WebSocket interface, SocketIO, ws, etc).

ericlu88:
Squashed.

PR #42, opened 2015-02-03
Made a python2.7.6 compatible example, using only core Python. Note that...

j-ruffer:
... the requests module would be more ideal, but it is not a core library, so I stuck to httplib.

This is a direct port of the HistoricalDataRequest.py example. I've tested this, though this should take some regression testing on other machines as well.

### DIFF
--- a/examples/python/python2/__main__.py
+++ b/examples/python/python2/__main__.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python2
+
+import argparse
+import json
+import sys
+import httplib
+import pprint
+
+
+data = {
+    "securities": ["IBM US Equity", "AAPL US Equity"],
+    "fields": ["PX_LAST", "OPEN", "EPS_ANNUALIZED"],
+    "startDate": "20120101",
+    "endDate": "20120301",
+    "periodicitySelection": "DAILY"
+}
+
+def request(args):
+    print args.host
+    print args.key
+    print args.cert
+    conn = httplib.HTTPSConnection(args.host,
+                                   key_file=args.key,
+                                   cert_file=args.cert)
+    headers = {'Content-Type': 'application/json'}
+    conn.set_debuglevel(4)
+    conn.request('POST', '/request/blp/refdata/HistoricalData', json.dumps(data), headers)
+    response = conn.getresponse()
+
+    print response.status, response.reason
+    pprint.pprint(json.loads(response.read()))
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Makes a historical data ' \
+                                     'request to the bloomberg http API for ' \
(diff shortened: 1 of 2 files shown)


examples/python/python2/__main__.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python2
+
+import argparse
+import json
+import sys
+import httplib
+import pprint
```
ericvw: nit: It may be good to sort the `import`'s alphabetically.  Not necessary though.

ericvw:
First, thanks for helping to contribute a Python example.  A few things that would be great to see changed before we land this.
1. Please rename the file to something like `HistoricalDataRequest.py` so that we may have other named examples in this new directory.
2. It may be great to add some Python version checking and error reporting when the version isn't supported by the example to reduce the element of surprise for users
3. I can't tell from the GitHub diff viewer, but ensure that the file has the executable bit set.

ericvw:
Another thing, we tend follow these [commit message guidelines](https://github.com/blog/926-shiny-new-commit-styles).  If you could update your commit message accordingly, that would be awesome.  We relax the 50 character summary limit to something reasonably longer - use your own judgement.

j-ruffer:
Sure thing, I'll update according to your suggestions.

j-ruffer:
Import dependencies exist here, so alphabetical ording wouldn't be ideal. However, I can do a version check so we can get an early exit from the example. I'll update my commit message appropriately.

ericvw:
So this does not work in `2.7.9`:

``` bash
$ ./HistoricalDataRequest.py -s http-api.openbloomberg.com
http-api.openbloomberg.com
client.key
client.crt
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)
```

We can't in good faith have code examples that are security risks :).  It turns out that `httplib` in `<= 2.7.9` disable server verification, which lends itself to potential man-in-the-middle attacks.

If we can use `urllib2` or `urllib` to accomplish this properly, that would be preferred.

ericvw:
Just checking in on the status of this PR?  @j-ruffer, any updates?

j-ruffer:
@ericvw Still in progress. Would you like me to close until I have a proper example?

ericvw:
It is OK to keep open.  I just wanted to touch base and probably will again next Monday with status.

ericvw:
@j-ruffer, an updates?

ericvw:
Haven't heard back from original requester.

j-ruffer:
I'm cool with this. My bad!

Sent from Bloomberg Professional for Android

----- Original Message -----
From: Eric N. Vander Weele 
At: 08-Nov-2015 15:35:56

Closed #42.

PR #44, opened 2015-02-04
Add subscription over long-polling

ericlu88:
This PR adds subscription functions over http long-polling.

Server usage:
- `subscribe`: **POST** only, with JSON subscription data as request
  body
- `poll`: **GET** only, with **pollid** specified as url arguments.
  - **pollid** is an integer that serve the function of data recovery
    and data acknowledgement. The client start the first poll request with
    any arbitrary integer. The client then increment the pollid by **1**
    in their next poll request in order to acknowledge receiving the
    previous data and is ready to receive new data. If for any reason the
    client doesn't receive the previous data, it will make another poll
    request with the previous pollid, and the server will resend the same
    data as it did in the previous response. **Note**: Once the pollid get
    incremented, there's no way going back to the previous value, since
    the incrementation is a signal to the server to drop old data.
- `unsubscribe`: **POST** only, with JSON correlation ids as request
  body, will only unsubscribe the specified subscriptions. The
  correlationIds validation logic is the same as implemented in
  `websocket`
- **NOTE**: The `subscription` feature only works in **HTTPS** mode,
  since we are using `client certificate` to map user with their
  session.

Some implementation notes:
- Add `map.ts` as generic map implementation that conforms to ES6 map
  interface
- Add `session.ts` to create stateful **session** for our http server,
  which holds user specific data(subscriptions and related).
- Add `session-store.ts` to hold `apiSession` and to handle session
  expiration
- Add `api-session.ts` middleware to handle session creation and
  retrieval. These middleware will only be executed for subscription
  related route, not request/response route.
- Add `historical-buffer-manager.ts` to handle multi-buffer functionality
- Add two test examples in `test` directory.


index.ts
```diff
@@ -35,6 +37,19 @@ server.use(requestHandler.elevateRequest);
...
 // Route
 server.post('/request/:ns/:svName/:reqName', requestHandler.onRequest);
+server.post('/subscribe', util.rejectHTTP,
+                          apiSession.createSession,
+                          apiSession.handleSession,
+                          requestHandler.onSubscribe);
+server.get('/poll', util.rejectHTTP,
+                    apiSession.handleSession,
+                    requestHandler.onPoll);
+server.get('/unsubscribe', util.rejectHTTP,
```
ericvw: Let's remove the HTTP GET `/unsubscribe` logic since GET's should theoretically by idempotent (i.e., not change the server-side state).

lib/config.ts
```diff
@@ -180,6 +180,26 @@ export function get(name: string): any
                     arg: 'websocket-ws-port'
                 },
             }
+        },
+        'longPoll': {
```
ericvw: For this line and the remaining part of this section of the config files, let's lower-case everything and use `-` where appropriate.  I think `longpoll` is acceptable.

lib/apisession/session-store.ts
```diff
@@ -0,0 +1,83 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import _ = require('lodash');
+import Session = require('./session');
+
+export = SessionStore;
+
+class SessionStore {
```
ericvw, line 8: This seems similar to `SubscriptionStore` that was introduced in #38.  To summarize our verbal discussion, let's provide a `Map` class that is is the same interfaces as the ES6 maps that can be extended.

lib/apisession/session.ts
```diff
@@ -0,0 +1,77 @@
...
+    private logger: bunyan.Logger;
+
+    // PUBLIC VARIABLES
+    public key: string;
+    public inUse: number = 0;
+    public activeSubscriptions = new SubscriptionStore();
+    public receivedSubscriptions = new SubscriptionStore();
+    public lastPollId: number = undefined;
+    public lastSuccessPollId: number = undefined;
+    public expired: boolean = false;
```
ericvw: Tidy up access control of these variables.

index.ts
ericlu88: Removed. Only `POST` is available for unsubscribe.

ericlu88:
As suggested, add `lib/util/strmap.ts` as a generic map implementation that implement the same interface as ES6 map. This removes the need of original `subscription-store.ts` since they are providing almost the same functionality. The revised `session-store.ts` extends this generic map and add only the auto expiration functionality, and the new `subscription-store.ts` add only the buffer related functions.

lib/middleware/util.ts
```diff
@@ -54,3 +54,16 @@ export function after(req: Interface.IOurRequest,
 {
     req.log.info({res: res}, 'Response sent.');
 }
+
+export function rejectHTTP(req: Interface.IOurRequest,
+                           res: Interface.IOurResponse,
+                           next: restify.Next): void
+{
+    if (!conf.get('https.enable')) {
```
yupinghu: This check seems odd -- we're not checking if the request itself is http (vs https), but rather whether the server is configured for https. Should this check req.isSecure() or some such?

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        delete this._map[mkey];
+    }
+
+    keys(): string[] {
+        var res: string[] = [];
+        for (var t in this._map) {
+            res.push(demangle(t));
+        }
+        return res;
```
ericvw: I wouldn't recommend using `in` because of the potential of iterating items on the prototype of `this._map`.  I functional approach that I would recommend:

``` TypeScript
return _.map(Object.keys(this._map), (k: string) => {
    return demangle(k);
});
```

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    values(): T[] {
+        var res: T[] = [];
+        for (var t in this._map) {
+           res.push(this._map[t]);
+        }
+        return res;
```
ericvw: I would recommend using `lodash` for a functional approach here as well.

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+           callbackFn.call(thisArg, this._map[mkey], demangle(mkey), this);
+        }
+    }
+
+    entries(): any[][] {
+        var res: any[][] = [];
+        for (var t in this._map) {
+            res.push([demangle(t), this._map[t]]);
+        }
+        return res;
```
ericvw: I would recommend using `lodash` for a functional approach here as well.

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+    values(): T[] {
+        var res: T[] = [];
+        for (var t in this._map) {
+           res.push(this._map[t]);
+        }
+        return res;
+    }
+
+    forEach(callbackFn: (val: T, key: string, map: StrMap<T>) => void, thisArg?: any): void {
+        for (var mkey in this._map) {
```
ericvw: I would recommend using `lodash` for a functional approach here as well (i.e., `_.forEach`)

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+    has(key: string): boolean {
+        return mangle(key) in this._map;
+    }
+
+    delete(key: string): void {
+        var mkey = mangle(key);
+        if (mkey in this._map) {
+            --this._size;
+        }
+        delete this._map[mkey];
```
ericvw: I would move this within the `if` block for clarity.  I know that it will work, but new users may find this pattern unusual.

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        this._map[mkey] = val;
+    }
+
+    get(key: string): T {
+        return this._map[mangle(key)];
+    }
+
+    has(key: string): boolean {
+        return mangle(key) in this._map;
```
ericvw: I would wrap the expression in parenthesis for clarity here to emphasize the boolean result is being returned.

lib/middleware/util.ts
ericvw: I agree that we should trust the underlying framework to tell us if it is a secure connection or not if possible.

lib/subscription/buffer.ts
```diff
@@ -0,0 +1,77 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import bunyan = require('bunyan');
+import conf = require('../config');
+
+export = Buffer;
+
+class Buffer<T> {
```
yupinghu: Some comments on the goals/design of this class would be great.

lib/subscription/subscription-buffer.ts
```diff
@@ -0,0 +1,63 @@
...
+    constructor( cid: number,
+                 security: string,
+                 fields: string[],
+                 maxBufferLength: number,
+                 options?: any) {
+        super(cid, security, fields, options);
+        this.buffer = new Buffer<Object>(maxBufferLength);
+    }
+
+    pushBuffer(data: any): void {
```
yupinghu: Should data be of type Object? (To match buffer's type.)

lib/subscription/subscription-buffer.ts
```diff
@@ -0,0 +1,63 @@
...
+            'data': buff[0],
+            'missed_ticks': buff[1]
+        };
+    }
+
+    getOldBuffer(): Object {
+        var buff = this.buffer.getOld();
+        return {
+            'correlationId': this.correlationId,
+            'data': buff[0],
```
yupinghu: I'm not sure I understand how buff[0] and buff[1] map to the data/missed_ticks; getOld (and getNew, for the previous function) return an array of multiple values so you should just return the entire array?

lib/subscription/subscription-buffer.ts
```diff
@@ -0,0 +1,63 @@
...
+import Subscription = require('./subscription');
+import Buffer = require('./buffer');
+
+export = SubscriptionWithBuffer;
+
+class SubscriptionWithBuffer extends Subscription {
+
+    private buffer: Buffer<Object>;
+
+    constructor( cid: number,
```
yupinghu: Remove space between ( and c

lib/util/strmap.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    values(): T[] {
+        var res: T[] = [];
+        for (var t in this._map) {
+           res.push(this._map[t]);
+        }
+        return res;
+    }
+
+    forEach(callbackFn: (val: T, key: string, map: StrMap<T>) => void, thisArg?: any): void {
```
ericvw: Let's make `key` and `map` optional.

lib/util/strmap.ts
ericvw: I'll update the style guide for the rules/conventions on this.

lib/util/buffer-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+        throw new Error();
+    }
+
+    // Flush the last data array from the history and move the entire history one step forward
+    // Subsequent push will be stored into a new array
+    flush(): void {
+        // Not Implemented Yet
+        throw new Error();
+    }
+}
```
ericlu88: @ericvw @yupinghu As discussed yesterday, I added this `BufferManager` class skeleton for review purpose. Once we settled on the interfaces, I can go ahead and implement it.

lib/util/buffer-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface IBufferedData<T> {
+    data: T[];
+    overflow: number;
+}
+
+export class BufferManager<T> {
+
+    // maxBuffLength: Maximum numbers of data item per level of buffer
+    // maxHistoryDepth: Maximum numbers of preserved buffer history
+    constructor(maxBuffLength: number, maxHistoryDepth: number) {
```
ericvw: Let's switch these two parameters around and have the maxBuffLength be optional with a predefined default.

lib/util/buffer-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+    // depth: Default is to retrieve the current buffer. The first history will be depth 1
+    // Return an object containing the data array and an integer for number of overlow items
+    get(depth?: number): IBufferedData<T> {
+        // Not Implemented Yet
+        throw new Error();
+    }
+
+    // Flush the last data array from the history and move the entire history one step forward
+    // Subsequent push will be stored into a new array
+    flush(): void {
```
ericvw: This could take `maxBufferLength` as an optional param for the new buffer.

ericvw: `flush` => `startNewBuffer`

lib/util/buffer-manager.ts
ericvw: `maxBufferLength`

lib/util/buffer-manager.ts
ericvw: Made my comments, let's see what @yupinghu has to say.

test/testLongPoll2.js
```diff
@@ -0,0 +1,131 @@
...
+                        longPolling();
+                    }
+                })
+                .catch(function(err) {
+                    console.log(err);
+                    process.exit();
+                });
+            }
+            else {
+                /*
```
ericvw: Can this commented out block be removed?

lib/util/buffer-manager.ts
ericlu88: I would argue these two arguments are equally weighted and be either all mandatory or all optional.

lib/util/buffer-manager.ts
ericlu88: I am thinking of all buffers are restricted to the same length `maxBufferLength`. I don't really see a usage of having each buffer has a custom length.

ericlu88:
Implement `HistoricalBufferManager` as discussed. Appreciate an eye on it. @ericvw @yupinghu

lib/util/map.ts
```diff
@@ -0,0 +1,65 @@
...
+        return this._size;
+    }
+
+    clear(): void {
+        this._size = 0;
+        this._map = {};
+    }
+
+    set(key: string|number, val: T): void {
+        if (!_.has(this._map, key.toString())) {
```
ericvw: Let's cache the result of `.toString()` since it is used twice in this function.

lib/util/map.ts
```diff
@@ -0,0 +1,65 @@
...
+    get(key: string|number): T {
+        return this._map[key.toString()];
+    }
+
+    has(key: string|number): boolean {
+        return _.has(this._map, key.toString());
+    }
+
+    delete(key: string|number): void {
+        if (_.has(this._map, key.toString())) {
```
ericvw: Let's cache the result of `.toString()` since it is used twice in this function.

ericvw:
As a side note, we found an issue (not in the diff yet) with `tslint 2.1.0` conflating multiple interfaces for `variables-before-functions` `member-ordering` checks.  I have PR palantir/tslint#302 to address this regression.

ericlu88:
Reimplement `HistoricalBufferManager` as discussed.

As a side note, I added test cases for this class since it is a perfect fit for unit testing. I used [Mocha](http://mochajs.org/) as the testing framework and [should.js](https://github.com/tj/should.js) as the assertion library. I choose these since they appear to be pretty popular(and pretty easy to use), but we can still try something else. If we don't want to deal with unit testing for this PR, I can strip it out and create another one, but I think this would be a good start for adding test coverage for our codebase.

Makefile
```diff
@@ -3,6 +3,7 @@ SRCS_JS_FROM_TS =
 
 SRCS_TS += index.ts
 SRCS_TS += $(shell find lib -type f -name '*.ts')
+SRCS_TS += $(shell find test -type f -name '*.ts')
```
ericvw: Can we make this that we only compile tests when executing the test target?  I can assist with the Makefile structure if it helps.

lib/middleware/api-session.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!apiSession) {
+        req.log.debug('No active apisession found.');
+        return next(new restify.BadRequestError('No active apisession found.'));
+    }
+    req.log.debug('apisession retrieved.');
+    ++apiSession.inUse;
+    req.apiSession = apiSession;    // Attach apisession to request
+
+    res.once('finish', (): void => {
+        --apiSession.inUse;
```
ericvw, line 59: I would assert that this never becomes negative.

lib/middleware/api-session.ts
```diff
@@ -0,0 +1,70 @@
...
+        req.log.debug({inuse: apiSession.inUse}, 'Response finished.');
+    });
+
+    // connection was terminated before response.end() was called
+    res.once('close', (): void => {
+        --apiSession.inUse;
```
ericvw, line 66: I would assert that this never becomes negative.

Makefile
ericlu88: I tried to achieve what you described over the weekend but no luck, will appreciate your help! @ericvw

ericlu88: Also, we should research integrating the Mocha tests with `Travis`

lib/subscription/buffer.ts
```diff
@@ -0,0 +1,77 @@
+/// <reference path='../../typings/tsd.d.ts' />
```
ericvw: I believe we can remove this file now in favor of `HistoricalBufferManager`

ericlu88: Yes, will remove this file and refactor relevant operations later.

Makefile
ericvw: I can do this since I already am somewhat familiar with it.  @ericlu88, it may be best to move the testing changes into a separate PR since this may take some time.  Is that cool with you?

ericlu88: @ericvw Sure. I will strip out the `Makefile` and changes related to unit tests into a separate PR. At least I am more confident for the `HistoricalBufferManager` now with the tests. :smile:

ericvw: I am all for tests - I just don't want to get distracted with getting that all hooked up right with this PR :).

lib/middleware/util.ts
```diff
@@ -21,7 +21,7 @@ export function getCert(req: Interface.IOurRequest,
                         next: restify.Next): void
 {
     // Get client certificate details(only in https mode)
-    if (conf.get('https.enable')) {
+    if (req.isSecure()) {
         req.clientCert = (<tls.ClearTextStream>req.connection).getPeerCertificate();
```
ericlu88, line 25: Has to change `restify` typing to expose this function. Depends on PR borisyankov/DefinitelyTyped#3626

ericlu88:
@ericvw @yupinghu Integrate `HistoricalBufferManager` into subscription functions, please give it a second go.

 I almost think `subscription-buffer` is not needed now since we push all the buffer related functions into `HistoricalBufferManager`. It only extends `subscription` by adding a `buffer` variable into it. The downside to pushing the buffer into `subscription` is to pollute the `websocket` use case with buffers. Appreciate any thoughts here.

ericlu88:
@yupinghu Good point. Implemented as a getter.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import assert = require('assert');
+import interfaces = require('../interface');
+
+// Internal class that holds the buffer array, a counter for overflow items and maxBufferLength
+class Buffer<T> {
+    data: T[];
+    overflow: number;
```
ericvw: Just a preference, but you can also move the initialization here for syntactic sugar:

``` typescript
data: T[] = [];
overflow: number = 0;
```

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import assert = require('assert');
+import interfaces = require('../interface');
+
+// Internal class that holds the buffer array, a counter for overflow items and maxBufferLength
+class Buffer<T> {
+    data: T[];
```
ericvw: These should probably be `private`.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+    overflow: number;
+    maxBufferLength: number;
+
+    get getObject(): interfaces.IBufferedData<T> {
```
ericvw: Let's move accessors after constructors.  I may get motivated to add a feature to `tslint` to check for this.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+        return { buffer: this.data.slice(),
```
ericvw: I would appreciate @yupinghu's comments as well.  I propose:

``` typescript
 return {
    buffer: this.data.slice(),
    overflow: this.overflow 
};
```

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export = HistoricalBufferManager;
+
+// This class creates a buffer manager which manage N buffers(N is taken from constructor parameter)
+class HistoricalBufferManager<T> {
+
+    // PRIVATE VARIABLES
+    private _buffers: Buffer<T>[];
+    private _currentIndex: number;
```
ericvw: Initial values can be set here as well for default values.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+            this._buffers.push(new Buffer<T>(this._defaultBuffLength));
+        }
+    }
+
+    // Push one data item to the current buffer
+    pushValue(data: T): void {
+        var currBuffer: Buffer<T> = this._buffers[this._currentIndex];
+        if (currBuffer.data.length === currBuffer.maxBufferLength) {
+            currBuffer.data.splice(0, 1);
+            ++currBuffer.overflow;
```
ericvw: I think it is common to have this be `currBuffer.overflow++` and it is more clear what is being incremented.

lib/util/historical-buffer-manager.ts
ericvw: Let's make this a method since it actually does some work to get something.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+        currBuffer.overflow = 0;  // Reset the overflow counter
+        return ret;
+    }
+
+    // Get buffered data array.
+    // depth: This optional parameter controls which level of history will be retrieved
+    // depth: Default is to retrieve the current buffer. The first history will be depth 1
+    // Return an object containing the data array and an integer for number of overflow items
+    getBuffer(depth: number = 0): interfaces.IBufferedData<T> {
+        assert(depth >= 0 && depth <= this._historyDepth - 1, 'Invalid depth parameter ' + depth);
```
ericvw: Not critical, but it is sometimes better to have these as separate asserts so that the assert message can be more descriptive why it fired.

lib/util/historical-buffer-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Flush the last data array from the history and move the entire history one step forward
+    // Subsequent push will be stored into a new array
+    // bufferLength: Optional parameter to control the length of the new buffer
+    // Return the buffered data of the most recent flushed array
+    startNewBuffer(bufferLength: number = this._defaultBuffLength): interfaces.IBufferedData<T> {
+        assert(bufferLength >= 1, 'Invalid depth parameter ' + bufferLength);
+        var currBuffer: Buffer<T> = this._buffers[this._currentIndex];
+        var ret: interfaces.IBufferedData<T> = currBuffer.getObject;
+        // Get the new current index
+        this._currentIndex = (this._currentIndex + this._historyDepth - 1) % this._historyDepth;
```
ericvw: It may be worth while making this calculation a private method taking in the parameter for the depth to pass in to be used in other places.

lib/util/historical-buffer-manager.ts
yupinghu: I think I like this style too, i.e. use initializers with the variable declarations. It's nice to have the defaults with the variables and it's nice when you have multiple constructors.

ericlu88: Got it. Will update accordingly.

lib/util/historical-buffer-manager.ts
yupinghu: As a style guide issue, I would propose that you use the style @ericvw proposes here in general (i.e. braces and properties don't share lines), with an exception for when you have just one property and you inline it all.

lib/interface.ts
```diff
@@ -3,16 +3,22 @@
...
 import BAPI = require('./blpapi-wrapper');
+import Session = require('./apisession/session');
 
 export interface IOurRequest extends restify.Request {
     clientCert?: any;
     blpSession: BAPI.Session;
+    apiSession?: Session;
+}
+
+export interface IBufferedData<T> {
```
ericvw, line 14: What's the motivation of putting `IBufferedData` here?  Why not have this be exported by `historical-buffer-manager`?

lib/util/historical-buffer-manager.ts
ericlu88: I am not sure if I want to mark the `data` field as private, especially this is an internal class. I can definitely add all access control over this class, just not sure if it is best practice to do it over internal class.

lib/util/historical-buffer-manager.ts
ericlu88: Updated as `getBufferedData()`

lib/util/historical-buffer-manager.ts
ericlu88: Like this too.

lib/middleware/api-session.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Use client cert fingerprint as session key
+    if (!apiSession) {
+        SESSIONSTORE.set(req.clientCert.fingerprint, new Session(req.blpSession));
+        req.log.debug('New apisession created.');
+    }
+
+    return next();
+}
+
+export function handleSession (req: Interface.IOurRequest,
```
ericvw: nit: `handleSession (` = > `handleSession(`

lib/middleware/api-session.ts
```diff
@@ -0,0 +1,73 @@
...
+import conf = require('../config');
+import Session = require('../apisession/session');
+import SessionStore = require('../apisession/session-store');
+import Interface = require('../interface');
+
+// GLOBAL
+var SESSIONSTORE = new SessionStore(conf.get('expiration'));
+
+// PUBLIC FUNCTIONS
+export function createSession (req: Interface.IOurRequest,
```
ericvw: nit: `createSession (` => `createSession(`

lib/util/historical-buffer-manager.ts
ericvw: I'll update the style guide on a section regarding object literals on this consensus.

lib/util/historical-buffer-manager.ts
ericlu88: Ok, will update accordingly.

lib/util/historical-buffer-manager.ts
ericvw: #60 has been created to track this.

lib/interface.ts
ericvw, line 14: I would recommend putting this in `historical-buffer-manager.ts` and other modules can re-export this interface.

lib/util/historical-buffer-manager.ts
ericvw: Note that there are a couple of places where multiples things are being checked in the same assert besides just this line.

lib/interface.ts
ericlu88, line 14: Having `historical-buffer-manager` to only contains one class enable me to use the `export =` syntax, which is much user friendly to use(don't need to call the class name again and let the user define what it is called). And since we have this interface file to put all interfaces that will be used across our application, I figure this could be a good place.

lib/util/historical-buffer-manager.ts
ericvw: Ahh, I see - I missed that this is internal.  I am OK with this.

lib/util/historical-buffer-manager.ts
yupinghu: But... but... preincrement is faster! ;)

lib/util/historical-buffer-manager.ts
yupinghu: I could go either way in this case; for bounds checking, I usually like just having an "out of bounds" assertion.

lib/interface.ts
ericvw, line 14: My concern is that the code that gives out objects adhering to this interface is not co-located with the interface itself.  It gives the impression that having the itnerface exported from `interface.ts` that `IBufferedData` can be used for multiple different APIs, when it is really just used for one.

lib/util/historical-buffer-manager.ts
ericvw: Only in languages that return iterators and deal with internal state :P

ericvw: Besides, V8 probably optimizes for this anyways :).

lib/subscription/subscription-store.ts
```diff
@@ -1,72 +1,47 @@
...
 import _ = require('lodash');
-import Subscription = require('./subscription');
+import Subscription = require('./subscription-buffer');
+import Map = require('../util/map');
+import interfaces = require('../interface');
 
 export = SubscriptionStore;
 
-class SubscriptionStore<T extends Subscription> {
+class SubscriptionStore extends Map<Subscription> {
```
ericvw: This shouldn't extend `Map`.  It would be preferable to use composition and have the `SubscriptionStore` have a member that is of type `Map<Subscription>`.  @yupinghu, thoughts?

lib/subscription/subscription-store.ts
```diff
@@ -1,72 +1,47 @@
...
-            }
-            delete this.subscriptions[value];
-            return true;
-        }
-        else if (typeof value === 'object') {
-            if (!this.has(value)) {
-                return false;
+    getAllOldBuffers(): Object[] {
+        var buffers: Object[] = [];
+        this.forEach((sub: Subscription, key: string, map: Map<Subscription>): void => {
```
ericvw: I would use `map` here to return `Object[]`.

lib/subscription/subscription-store.ts
```diff
@@ -1,72 +1,47 @@
...
-        else if (typeof value === 'object') {
-            return _.has(this.subscriptions, value.correlationId);
+    startAllNewBuffers(): Object[] {
+        var buffers: Object[] = [];
+        // Check if we have new data for any subscriptions
+        var hasNewData: boolean = _.some(this._map, (sub: Subscription): boolean => {
+            return !sub.buffer.isEmpty();
+        });
+        if (hasNewData) { // Start a new buffer for every subscription and return back the new data
+            this.forEach((sub: Subscription, key: string, map: Map<Subscription>): void => {
```
ericvw: I would use `map` here to return `Object[]`.

lib/interface.ts
ericlu88, line 14: I see javascript interface is a contract across files, so the impression is any user code who expect `IBufferedData` will import the interface from `interface.ts`, but the data won't necessary come from `historical-buffer-manager`.

lib/util/historical-buffer-manager.ts
ericlu88: I lean toward post increment too here for better readability

lib/subscription/subscription-store.ts
ericlu88: Is there a way to tell `_.map()` not add an element to the array?(ie, the return array's length is less by or equal to the original collection)

lib/subscription/subscription-store.ts
yupinghu: I generally agree that I usually think of it as being a "has a" relationship, not "is a". That is, I think of SubscriptionStore as having a map that it uses to store its internal state, rather than being a map. The main downside is that you'll need to define a bunch of functions to just delegate to the map member, though.

ericlu88: Or, mark the map as **public**?

yupinghu: Yeah you could do that too. I tend to prefer not to do that for encapsulation but it may make sense here, I'd have to look at it closer at how it's used.

ericlu88: Yeah Let me examine all usage and see if it make sense to wrap the needed operations.

lib/subscription/subscription-store.ts
yupinghu: Use _.filter to remove the guys who shouldn't be added to the output array, and then chain _.filter into _.map.

ericlu88:
@yupinghu Implement the `getIndex()` as suggested.

lib/interface.ts
ericvw, line 14: I'm OK with this since I can't come up with a strong argument for an alternative solution.

typings/lodash/lodash.d.ts
```diff
@@ -2641,6 +2641,15 @@ declare module _ {
...
 
+    interface LoDashObjectWrapper<T> {
+        /**
+        * @see _.filter
+        **/
+        filter<T extends {}>(
+            callback: ObjectIterator<T, boolean>,
+            thisArg?: any): LoDashObjectWrapper<T>;
+    }
+
```
ericlu88: Has to extend the `LoDashObjectWrapper` in order to chain `filter` function, not sure why the lodash typing file doesn't expose it. I've created a PR for this change(borisyankov/DefinitelyTyped#3631)

lib/subscription/subscription.ts
```diff
@@ -1,19 +1,25 @@
...
 
 import BAPI = require('../blpapi-wrapper');
+import BufferManager = require('../util/historical-buffer-manager');
 
 export = Subscription;
 
 class Subscription extends BAPI.Subscription {
 
     correlationId: number;
+    buffer: BufferManager<Object>;
```
ericvw: `buffer: BufferManager<Object> = null;` would preferred instead of `undefined`.

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    get receivedSubscriptions(): Map<Subscription> {
+        return this._receivedSubscriptions;
+    }
+
+    // PUBLIC VARIABLES
+    public inUse: number = 0;
+    public lastPollId: number = undefined;
+    public lastSuccessPollId: number = undefined;
```
ericvw: s/`undefined`/`null` for these two lines.  If any of these are read-only, I would recommend that we make these property getters.

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    isExpirable(): boolean {
+        return !this.inUse;
+    }
+
+    renew(): void {
+        this._seconds = curSeconds();  // Record the renewed time
+    }
+
+    // PRIVATE FUNCTIONS
```
ericvw: Let's move private functions before `// CREATORS`

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+            }
+            this._logger.debug('Unsubscribed all active subscriptions.');
+        }
+        this.clear();
+        this._expired = true;
+        this._logger.debug('Session expires.');
+        return true;
+    }
+
+    isExpirable(): boolean {
```
ericvw: Let's move public `// ACCESSORS` after public `// MANIPULATORS`

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+        return !this.inUse;
+    }
+
+    renew(): void {
+        this._seconds = curSeconds();  // Record the renewed time
+    }
+
+    // PRIVATE FUNCTIONS
+    private clear(): void {
+        if (this._receivedSubscriptions.size) {
```
ericvw: I think you can just call `forEach` here without checking `.size` and it it should be a no-op.  It would make this section a bit more elegant.

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+            var subscriptions = this._receivedSubscriptions.values();
+            subscriptions.forEach((sub: Subscription): void => {
+                sub.removeAllListeners();
+            });
+        }
+        this._activeSubscriptions.clear();
+        this._receivedSubscriptions.clear();
+        this.lastPollId = this.lastSuccessPollId = undefined;
```
ericvw: s/`undefined`/`null`/

lib/apisession/session.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+function curSeconds(): number
```
ericvw: Let's move this before the class so we know where this function got introduced when reading the class.

lib/apisession/session.ts
ericlu88: No these are writable by `request-handler`. I will mark them as `null` for default case.

lib/middleware/request-handler.ts
```diff
@@ -138,3 +130,371 @@ export function onRequest(req: Interface.IOurRequest,
...
+                subscriptions.forEach((s: Subscription): void => {
+                    s.removeAllListeners();
+                    req.apiSession.receivedSubscriptions.delete(s.correlationId);
+                });
+                req.log.debug('Unsubscribed all active subscriptions');
+                return Promise.resolve();
+            }
+        })
+        .then(next)
+        .catch((err: Error): Promise<any> => {
```
ericvw: `Promise<void>`?

ericlu88: I guess it really should be `<any>` since this is the type definition for `next`:

``` typescript
interface Next {
    (err?: any): any;
  }
```

ericvw:
lgtm.  I will defer to @yupinghu's sign-off and will let him merge.

PR #45, opened 2015-02-04
Add note about OSX 'curl' incompatibility to examples.

kpfleming:
Add a notice to the 'curl'-based examples about the current incompatibility with 'curl' provided by OSX 10.9+.

### DIFF
--- a/examples/shell/FieldInfo.sh
+++ b/examples/shell/FieldInfo.sh
@@ -2,6 +2,10 @@
 
 # usage: ./FieldInfo.sh <request.json>
 
+# NOTE: This example program currently fails on Mac OSX 10.9 or later, due to the inability
+# of 'curl' supplied in OSX to accept certificates and keys through command-line arguments.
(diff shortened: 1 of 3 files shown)


ericvw:
First, thanks for the contributing to the project!

I would prefer that we roll this comment into one `README.md` in the `examples/shell` directory; it would provide a single location for the notice and reduce the burden of keeping the comment in-sync across multiple files.  Also, I would note for the Homebrew/MacPorts builds that it they should also be built with OpenSSL support - Homebrew's formula for `curl` defaults to building against the Secure Transport library by default.

kpfleming:
Fair enough all around. I may try to do that tomorrow, although I have very little time left before traveling again, and was trying to provide _something_ for the teams that will be using this stuff this weekend.

ericvw:
Just checking in on the status of this PR.  @kpfleming, any updates?

ericvw:
@kpfleming, any updates?

kpfleming:
No, sorry, my 'spare time' has disappeared. Feel free to close this.

PR #46, opened 2015-02-05
Upgrade tslint to 2.1.0

ericlu88:
This PR upgrades `tslint` to version 2.1.0. 

We found a bug with tslint(palantir/tslint#292) that prevents us doing object literals. We have to temporarily disable this check for our one usage, but will remove it once tslint fix the issue and bump up the version.


ericvw:
lgtm.  I will let @yupinghu if he is cool with everything.

yupinghu:
This is so dumb, that this is necessary. :) But lgtm.

PR #47, opened 2015-02-06
add more tslint checks

ericvw:
- enforece PascalCased class names
- enforece properly checks when using `for ... in` constructs
- interfaces names must begin with capital `I`
- no unused expressions (i.e., expressions that contain side-effects)
- enforce type definitions in function parameters, properties for
  interface declarations, and member variable declarations

Affected files by the added checks have been updated appropriately.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -191,7 +191,7 @@ export function get(name: string): any
     // Build options object
     // Bunyan logger options
     // Override default bunyan response serializer
-    bunyan.stdSerializers['res'] = function(res): any {
+    bunyan.stdSerializers['res'] = function(res: any): any {
         if (!res || !res.statusCode) {
             return res;
         }
@@ -201,7 +201,7 @@ export function get(name: string): any
         };
     };
     // Add client cert serializer
-    bunyan.stdSerializers['cert'] = function(cert): any {
+    bunyan.stdSerializers['cert'] = function(cert: any): any {
         return cert && {
             CN: cert.subject.CN,
             fingerprint: cert.fingerprint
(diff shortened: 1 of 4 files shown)


lib/middleware/request-handler.ts
```diff
@@ -87,7 +87,7 @@ export function elevateRequest (req: Interface.IOurRequest,
         });
     };
 
-    res.sendError = function(err, where, reason): Promise<void> {
+    res.sendError = function(err: any, where: string, reason: any): Promise<void> {
```
ericlu88, line 90:
``` typescript
err: Error, where: string, reason: string
```

lib/config.ts
```diff
@@ -191,7 +191,7 @@ export function get(name: string): any
     // Build options object
     // Bunyan logger options
     // Override default bunyan response serializer
-    bunyan.stdSerializers['res'] = function(res): any {
+    bunyan.stdSerializers['res'] = function(res: any): any {
```
ericlu88, line 194: `res: interface.IOurResponse` ?

lib/middleware/request-handler.ts
ericlu88, line 90: I will vote for getting rid of `res.sendError` and let `restify` handle the sending of errors.(The method has 0 usage in our codebase now).

ericvw, line 90: I concur.

lib/config.ts
ericvw, line 194: Decided to keep with `any` because it can by `any`thing.

lib/middleware/request-handler.ts
ericvw, line 90: Let's leave it for now for the purposes of this PR and then in a follow-up remove this.

ericvw:
Based on discussion, this is ready for merge if approved by reviewers.

ericlu88:
LGTM.

PR #48, opened 2015-02-06
update style guide regarding expressions in return statements

ericvw:


ericvw:
Just rebased against latest `develop` branch.

doc/style-guide.md
```diff
@@ -44,3 +45,52 @@ quaility.
...
+  }
+
+  // good
+  function foo(a: number) {
+      return (a > 1);
+  }
+
+  // good
+  function foo(a: number) {
+      return (a > 1) ? true : false;
```
yupinghu: Maybe change the return values to non-boolean. In this example, I agree with the stylistic issue here of where the parens go, but I'd complain that you shouldn't use a ternary operator because you should just be returning (a > 1). :)

doc/style-guide.md
```diff
@@ -44,3 +45,52 @@ quaility.
     character limit.
 
 **[top](#table-of-contents)**
+
+## Expressions
+
+  - Expressions in return statements should be surrounded by parenthesis for clarity.
```
yupinghu, line 51: Is there better wording to indicate we mean expressions that aren't just a value? Technically "a" is an expression, right? Or does javascript have different terminology for that?

ericvw, line 51: That is the note below about expressions containing one operand.  Is that not sufficient?

yupinghu, line 51: Yeah I'm mostly being pedantic here. :) I just wanted a word that captures both `return a` and `return !a` (the latter being what I think of when I hear "one operand"), but can't think of something beyond "Complex expressions in return statements..." which isn't great either.

doc/style-guide.md
ericvw: Is `return (a > 1) ? -1 : 0;` acceptable?

yupinghu: Yeah perfect. :)

doc/style-guide.md
ericvw, line 51: Since we cannot come up with something better, let's just leave it as is for now until we can come up with something better.  Cool?

yupinghu, line 51: Yeah that's fine by me.

ericvw:
Updated PR incorporating changes to the ternary example to return integers instead of booleans.

yupinghu:
LGTM

PR #49, opened 2015-02-07
added example for python3

fyquah:


ericvw:
Hey @fyquah95, thanks for your contribution.  The only feedback I have before we merge this PR is if you could amend the commit to rename the file to be under `examples/python/python3/HistoricalDataRequest.py`.  For consistency, we are doing something similar PR #42.  Thanks!

fyquah:
Hi @ericvw , I made the mentioned amendments!

examples/python/python3/HistoricalDataRequest.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+
+# usage: python3 HistoricalDataRequest.py <host-ip>
+
+import argparse
+import json
+import ssl
+import sys
+import urllib.request
```
ericvw, line 9: This can just be `import urllib`.

ericvw:
There is a minor in-line comment that would be great to change and if you wouldn't mind updating the commit summary to be along the lines of "add Python3 HistoricalDataRequest example" so that it is a bit more descriptive.  I'll go ahead and merge once those changes are done.  Thanks again for helping out with the project.  Every bit helps :).

examples/python/python3/HistoricalDataRequest.py
fyquah, line 9: I have tried `import urllib` but got the error `AttributeError: 'module' object has no attribute 'request'`. `import urllib.request` worked well tho.

fyquah:
Renamed the commit message to something more informative! (Similiar to those for the python example)

examples/python/python3/HistoricalDataRequest.py
ericvw, line 9: I didn't have Python3 installed, but looking at the docs for `urllib`, it looks like they broke out this module into smaller sub-modules.  Thanks for taking the time for double-check and verify this.

ericvw:
Thank you for contributing to the Python examples!

PR #50, opened 2015-02-07
Update HistoricalDataRequest.js

davidawad:
The previous code doesn't work out of the box and assumes you're running some bloomberg related code on localhost

### DIFF
--- a/examples/node/HistoricalDataRequest.js
+++ b/examples/node/HistoricalDataRequest.js
@@ -2,7 +2,7 @@
 var https = require('https');
 var fs = require('fs');
 
-var host = process.argv[2] || '127.0.0.1';
+var host = "http-api.openbloomberg.com";
 var port = 443
 
 var options = {


examples/node/HistoricalDataRequest.js
```diff
@@ -2,7 +2,7 @@
 var https = require('https');
 var fs = require('fs');
 
-var host = process.argv[2] || '127.0.0.1';
+var host = "http-api.openbloomberg.com";
```
ericvw, line 5: Can you update this to be:

``` javascript
var host = process.argv[2] || "http-api.openbloomberg.com";
```

please?

It would keep with the original intent of the example with just updating the default host.  Thanks!

ericvw:
As a general comment, please amend the commit message to be more descriptive about the update.  A good commit message summary may be `default HistoricalDataRequest.js host to http-api.openbloomberg.com`.  Thanks!

ericvw:
I amended the commit under 066339bfd654addf79a555ae61e70d8f60471e1c and landed it on `develop`.  @DavidAwad has still been left intact as the original author.  Thanks for the contribution!

PR #61, opened 2015-02-10
update style guide addressing object/array literals

ericvw:


doc/style-guide.md
```diff
@@ -91,6 +92,103 @@ quaility.
...
+  - Object literals
+    - Use curly braces `{}` instead of `new Object()`.
+    - The opening curly brace (i.e., `{`) must be on the same line as the declaration, expression,
+      or statement.
+    - The closing curly brace (i.e., `}`) must be on its own line aligned at the same indentation
+      level as line with the opening curly brace.
+    - Each `property: value` pair must be on its own line followed by a comma `,` indented by the
+      [indentation](#indentation) style specified within this document.
+    - The final property/value pair *should not* be followed by a comma `,`.
+    - If a object contains at most one property/value pair, it acceptable to have the
```
yupinghu: "If a**n** object ... it **is** acceptable ..."

doc/style-guide.md
```diff
@@ -91,6 +92,103 @@ quaility.
 
 **[top](#table-of-contents)**
 
+## Object and Array Literals
+
+  - Object literals
+    - Use curly braces `{}` instead of `new Object()`.
+    - The opening curly brace (i.e., `{`) must be on the same line as the declaration, expression,
```
yupinghu, line 99: If there's a graceful way to do so, it'd be good to call out that there's an exception to this rule (possibly in the exception below). Maybe have this and the following rule say that if there are properties? (Since the examples basically say that if there are no properties, `foo = {}` is the right way to do it.)

doc/style-guide.md
```diff
@@ -91,6 +92,103 @@ quaility.
...
+    // bad
+    var arr = [
+        1,
+        2,
+    ]
+
+    // good
+    var arr = [];
+
+    // good
```
yupinghu: Is this good, or OK?

doc/style-guide.md
```diff
@@ -91,6 +92,103 @@ quaility.
...
+  - Array literals
+    - Use brackets `[]` instead of `new Array()`.
+    - The opening bracket (i.e., `[`) must be on the same line as the declaration, expression, or
+      statement.
+    - The closing bracket (i.e., `]`) must be on its own line aligned at the same indentation level
+      as line with the opening bracket.
+    - Each array element must be on its own line followed by a comma `,` indented by the
+      [indentation](#indentation) style specified within this document.
+    - The final array element *should not* be followed by a comma `,`.
+    - It is acceptable to have an array literal on one line if and only if it fits within
```
ericlu88: I will vote for this as the default style for array literal(if not the only option) since it is the most common use case. One line per element seems a bit weird to me. @yupinghu, thoughts?

doc/style-guide.md
```diff
@@ -91,6 +92,103 @@ quaility.
...
+    // good
+    var arr = [ 1, 2, 3, 4, 5 ];
+
+    // good
+    var arr = [
+        1
+    ];
+
+    // good
+    var arr = [
```
ericlu88, line 196: This in my opinion should be the default option for multi-dimensional array(2)

ericvw:
Updated PR based on most recent feedback.

doc/style-guide.md
```diff
@@ -91,6 +92,117 @@ quaility.
...
+        bar: 2
+    };
+    ```
+
+  - Array literals
+    - Use brackets `[]` instead of `new Array()`.
+    - It is preferable to have an array literal on one line if and only if it fits within
+      [line length](#line-length) style specified within this document with a space after the
+      opening `[` and before the closing `]`.  If the array elements are of varying length, please
+      use your own judgement as to whether list each element on separates as defined in the
```
yupinghu: "... whether **to** list ... on separate **lines** ..."

ericvw:
Updated PR based on most recent feedback.

PR #64, opened 2015-02-11
Added Instrument lookup

thmiceli:
I don't have working examples for Curve and Government lookup requests yet but I should be able to come up with some. 
There isn't an example for FieldSearch either. I added the service to the map because Yu Ping mentioned it.
We should probably try to have at least one example for each service that is supported.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -71,13 +71,23 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+// Map of HTTP request names to the BLAPI request and response names.
+var NAME_MAP: { [s: string]: { reqName: string; resName: string; } } = {
+    'FieldInfo':        { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'FieldSearch':      { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'InstrumentLookup': { reqName: 'instrumentListRequest',
+                          resName: 'InstrumentListResponse' },
+    'CurveLookup':      { reqName: 'curveListRequest',
+                          resName: 'CurveListResponse' },
+    'GovernmentLookup': { reqName: 'govtListRequest',
+                          resName: 'GovtListResponse' }
+};
+
 // ANONYMOUS FUNCTIONS
 function isObjectEmpty(obj: Object): boolean {
     return (0 === Object.getOwnPropertyNames(obj).length);
-}
-
-function reqName2RespName(name: string): string {
-    return name.replace(/^Field\w+/, 'field') + 'Response';
 }
 
 function subscriptionsToServices(subscriptions: Subscription[]): string[] {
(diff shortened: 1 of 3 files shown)


lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+// Map of HTTP request names to the BLAPI request and response names.
```
ericlu88, line 74: BLPAPI

lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
...
+var NAME_MAP: { [s: string]: { reqName: string; resName: string; } } = {
```
yupinghu, line 75: This is a good test case for the style guide. :) @ericlu88 @ericvw what do you think of the formatting for NAME_MAP's value? It looks OK but it technically violates the current version of the style guide.

Also, we may want to consider a section on how to format type definitions.

lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
...
+    'FieldInfo':        { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'FieldSearch':      { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'InstrumentLookup': { reqName: 'instrumentListRequest',
+                          resName: 'InstrumentListResponse' },
+    'CurveLookup':      { reqName: 'curveListRequest',
```
yupinghu, line 82: Do we want this to be alphabetical?

lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
...
+                          resName: 'CurveListResponse' },
+    'GovernmentLookup': { reqName: 'govtListRequest',
+                          resName: 'GovtListResponse' }
+};
+
```
ericlu88, line 87: Seems to missing `HistoricalData` & `ReferenceData` in this list(The other two I am aware of). I am not sure if there's a complete list of request names in the api doc though.

lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
...
+                          resName: 'fieldResponse' },
+    'FieldSearch':      { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'InstrumentLookup': { reqName: 'instrumentListRequest',
+                          resName: 'InstrumentListResponse' },
+    'CurveLookup':      { reqName: 'curveListRequest',
+                          resName: 'CurveListResponse' },
+    'GovernmentLookup': { reqName: 'govtListRequest',
+                          resName: 'GovtListResponse' }
+};
```
yupinghu, line 86: Need to add entries for //blp/refdata requests. I believe HistoricalData and ReferenceData are the only ones we generally have tested, although there's no real reason the rest shouldn't work. The reqName and resName should just be FooRequest and FooResponse for all refdata requests.

thmiceli, line 86: Yes it looks like I missed these two services.

lib/blpapi-wrapper.ts
```diff
@@ -71,15 +71,25 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+// Map of HTTP request names to the BLAPI request and response names.
+var NAME_MAP: { [s: string]: { reqName: string; resName: string; } } = {
+    'FieldInfo':        { reqName: 'FieldInfoRequest',
+                          resName: 'fieldResponse' },
+    'FieldSearch':      { reqName: 'FieldInfoRequest',
```
ericlu88, line 78: `FieldSearchRequest`

thmiceli:
Replaced with PR #74

Issue #18, opened 2015-01-13
Testing infrastructure


ericvw:
We need this.
- We need to decided upon some testing infrastructure (i.e., mocha?)
- How can we stub the BLP API?
- Can the testing infrastructure work with TravisCI?

ericvw:
@ericlu88, we can share this issue and use it for tracking progress/updates on the testing infrastructure.

ericlu88:
Sure, forgot we have an issue for this.

ericvw:
Landed #69 to provide initial infrastructure support for this.

ericvw:
The first and third bullet points have been completed up until this point.  The remaining portion of this issue will now be dedicated towards stubbing the `blpapi-wrapper.ts` for testing purposes.

ericlu88:
After #94 is merged, the left over of this issue is to enhance the testing infrastructure with subscription related functionality.

ericvw:
Closed by #116

PR #70, opened 2015-02-11
Add test-historical-buffer-manager using mocha and should

ericlu88:
This PR adds a unit test for `historical-buffer-manager` using `mocha` as the test framework and `should.js` as the assertion library. 

This change belongs to #18


ericvw:
Can you update the `Makefile` to also lint the tests?

ericvw:
Also, I don't thinks it is necessary to mention the library dependencies for testing in the commit summary.  However, I don't feel strongly about it.

ericvw:
lgtm - please squash.  I'll defer to @yupinghu for the final merge.

PR #73, opened 2015-02-13
Convert to new URL scheme.

### LINKED ISSUE #59: Lock-in URL scheme for the HTTP API
This is to come to a consensus among the URL scheme and design of the HTTP API.  Once we understand how all the services work that is being addressed in #15, #54, #55, #56, #57, #58 we decide upon what the final HTTP API should look like with confidence.

yupinghu:
Closes #59.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -3,10 +3,13 @@
 import restify = require('restify');
 import Promise = require('bluebird');
 import bunyan = require('bunyan');
+import validator = require('express-validator');
 import conf = require('./lib/config');
+import Interface = require('./lib/interface');
 import blpSession = require('./lib/middleware/blp-session');
 import requestHandler = require('./lib/middleware/request-handler');
 import util = require('./lib/middleware/util');
+import requestValidator = require('./lib/middleware/request-validator');
 import sio = require('socket.io');
 import webSocket = require('ws');
 import webSocketHandler = require('./lib/websocket/websocket-handler');
@@ -29,28 +32,35 @@ server.use(restify.gzipResponse());
 server.use(restify.fullResponse());
 server.use(restify.throttle(conf.get('throttleOptions')));
 server.use(restify.requestLogger());
-server.use(restify.queryParser());
+server.use(restify.queryParser({ mapParams: false }));
+server.use(validator());
 server.use(util.getCert);
 server.use(util.log);
 server.use(blpSession.getSession);
 server.use(requestHandler.elevateRequest);
 
-// Route
-server.post('/request/:ns/:svName/:reqName',
+// Routing
+
+// `/request` handles request/response services. There is one operation on these services, namely
+// "get me some data". In some sense this ought to be a GET but because the request data is large
+// we put it in the body.
+server.post('/request',
+            requestValidator.requireQueryParams(['ns', 'service', 'type']),
             requestHandler.onRequest);
-server.post('/subscribe',
+
+// `/subscription` handles subscription services. There are two things you can do with
+// subscriptions: Modify your subscriptions and poll for subscription data. These are done via POST
+// and GET respectively.
+server.post('/subscription',
             util.rejectHTTP,
-            apiSession.createSession,
+            requestValidator.requireChoiceQueryParam('action', ['start', 'stop']),
             apiSession.handleSession,
-            requestHandler.onSubscribe);
-server.post('/unsubscribe',
-            util.rejectHTTP,
-            apiSession.handleSession,
-            requestHandler.onUnsubscribe);
-server.get('/poll',
+            requestHandler.onChangeSubscriptions);
+server.get('/subscription',
            util.rejectHTTP,
+           requestValidator.requireIntQueryParams(['pollid']),
            apiSession.handleSession,
-           requestHandler.onPoll);
+           requestHandler.onPollSubscriptions);
 
 // Listen
 server.listen(conf.get('port'));
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -69,13 +69,27 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+// Mapping of request types to response names to listen for.
+// The strings are taken from section A of the BLPAPI Developer's Guide, and are organized by
+// service.
+var REQUEST_TO_RESPONSE_MAP: { [index: string]: string; } = {
+    // //blp/refdata
+    'HistoricalDataRequest': 'HistoricalDataResponse',
+    'IntraDayTickRequest':   'IntraDayTickResponse',
+    'IntraDayBarRequest':    'IntraDayBarResponse',
+    'ReferenceDataRequest':  'ReferenceDataResponse',
+    'PortfolioDataRequest':  'PortfolioDataResponse',
+    'BeqsRequest':           'BeqsResponse',
+
+    // //blp/apiflds
+    'FieldInfoRequest':              'fieldResponse',
+    'FieldSearchRequest':            'fieldResponse',
+    'CategorizedFieldSearchRequest': 'categorizedFieldResponse'
+};
+
 // ANONYMOUS FUNCTIONS
 function isObjectEmpty(obj: Object): boolean {
     return (0 === Object.getOwnPropertyNames(obj).length);
-}
-
-function reqName2RespName(name: string): string {
-    return name.replace(/^Field\w+/, 'field') + 'Response';
 }
 
 function subscriptionsToServices(subscriptions: Subscription[]): string[] {
@@ -272,17 +286,19 @@ export class Session extends events.EventEmitter {
         }).nodeify(cb);
     }
 
-    request(uri: string, name: string, request: any, callback: IRequestCallback): void {
+    request(uri: string, requestName: string, request: any, callback: IRequestCallback): void {
         this.validateSession();
 
         var correlatorId = this.nextCorrelatorId();
         this.requests[correlatorId] = callback;
 
         this.openService(uri).then((): void => {
-            var responseEventName = reqName2RespName(name);
-            var requestName = name + 'Request';
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
+            if (!(requestName in REQUEST_TO_RESPONSE_MAP)) {
+                return callback(new Error('Unknown request type ' + requestName));
+            }
+            var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
             this.session.request(uri, requestName, request, correlatorId);
             this.listen(responseEventName,
                         correlatorId,
@@ -372,5 +388,3 @@ export class Session extends events.EventEmitter {
         }).valueOf());
     }
 }
-
-
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -242,8 +242,10 @@ export function get(name: string): any
     };
 
     // Restify bodyParser plugin options
-    otherConf['bodyParserOptions']
-        = { maxBodySize: convictConf.get('maxBodySize') };
+    otherConf['bodyParserOptions'] = {
+        maxBodySize: convictConf.get('maxBodySize'),
+        mapParams: false
+    };
 
     // Restify throttle plugin options
     otherConf['throttleOptions'] = {
--- a/lib/middleware/api-session.ts
+++ b/lib/middleware/api-session.ts
@@ -12,28 +12,6 @@ import Interface = require('../interface');
 var SESSIONSTORE = new SessionStore(conf.get('expiration'));
 
 // PUBLIC FUNCTIONS
-export function createSession(req: Interface.IOurRequest,
-                              res: Interface.IOurResponse,
-                              next: restify.Next): void
-{
-    // Check to make sure client cert is present
-    if (!_.has(req, 'clientCert')) {
-        req.log.error('No client certificate found.');
-        return next(new restify.InternalError('No client certificate found.'));
-    }
-
-    // Try to load the session associate with the fingerprint
-    var apiSession: Session = SESSIONSTORE.get(req.clientCert.fingerprint);
-    // Create a new session if no session found
-    // Use client cert fingerprint as session key
-    if (!apiSession) {
-        SESSIONSTORE.set(req.clientCert.fingerprint, new Session(req.blpSession));
-        req.log.debug('New apisession created.');
-    }
-
-    return next();
-}
-
 export function handleSession(req: Interface.IOurRequest,
                               res: Interface.IOurResponse,
                               next: restify.Next): void
@@ -46,10 +24,18 @@ export function handleSession(req: Interface.IOurRequest,
 
     // Try to load the session associate with the fingerprint
     var apiSession: Session = SESSIONSTORE.get(req.clientCert.fingerprint);
-    // Return error if no session found
+
     if (!apiSession) {
-        req.log.debug('No active apisession found.');
-        return next(new restify.BadRequestError('No active apisession found.'));
+        // If this is a new subscription, create the session.
+        if (req.method === 'POST' && req.params.action === 'start') {
+            apiSession = new Session(req.blpSession);
+            SESSIONSTORE.set(req.clientCert.fingerprint, apiSession);
+            req.log.debug('New apisession created.');
+        } else {
+            // When unsubscribing, return error if no session found.
+            req.log.debug('No active apisession found.');
+            return next(new restify.BadRequestError('No active apisession found.'));
+        }
     }
     req.log.debug('apisession retrieved.');
     ++apiSession.inUse;
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -1,9 +1,11 @@
 /// <reference path='../../typings/tsd.d.ts' />
 
+import assert = require('assert');
 import Promise = require('bluebird');
 import _ = require('lodash');
 import restify = require('restify');
 import bunyan = require('bunyan');
+import util = require('util');
 import Interface = require('../interface');
 import conf = require('../config');
 import Subscription = require('../subscription/subscription');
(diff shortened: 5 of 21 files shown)


yupinghu:
This is only the first version. At minimum, the docs and examples need to be fixed to match the new url scheme. Also, I want to add some query string validation.

lib/middleware/request-handler.ts
```diff
@@ -494,3 +476,28 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
+
+    switch (req.params.action) {
+        case 'subscribe':
+            return onSubscribe(req, res, next);
+        case 'unsubscribe':
+            return onUnsubscribe(req, res, next);
+    }
+    var errorString = 'Invalid realtime action: ' + req.params.action;
+    req.log.error(errorString);
+    return next(new restify.InternalError(errorString));
```
ericlu88: I think a `BadRequestError(400)` might be more appropriate here.

lib/middleware/request-handler.ts
```diff
@@ -191,11 +191,13 @@ export function onRequest(req: Interface.IOurRequest,
         return next(new restify.InternalError('Error not find blpSession.'));
     }
 
+    // TODO validate query params.
```
ericlu88: I was gonna add some comments about validating the query parameters before seeing this **TODO** item. :smile:

lib/middleware/api-session.ts
```diff
@@ -48,8 +26,15 @@ export function handleSession(req: Interface.IOurRequest,
     var apiSession: Session = SESSIONSTORE.get(req.clientCert.fingerprint);
     // Return error if no session found
```
ericlu88: This comment becomes invalid

lib/blpapi-wrapper.ts
```diff
@@ -69,15 +69,30 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+type StringToStringMap = {
+    [s: string]: string;
+};
+
+var REQUEST_TO_RESPONSE_MAP: StringToStringMap = {
```
ericlu88: I would vote for using a direct type definition instead of type alias for simplicity:

``` typescript
var REQUEST_TO_RESPONSE_MAP: {[index: string]: string}
```

index.ts
```diff
@@ -29,28 +30,32 @@ server.use(restify.gzipResponse());
...
 
-// Route
-server.post('/request/:ns/:svName/:reqName',
+// Routing
+
+// `/request` handles request/response services. There is one operation on these services, namely
+// "get me some data". In some sense this ought to be a GET but because the request data is fairly
+// large we use a POST.
+server.post('/request',
             requestHandler.onRequest);
```
ericvw, line 49: nit: Since this can be joined with the previous line, I would join them together.  However, I understand why it it is aligned on its own line because of the statements that follow.

index.ts
```diff
@@ -29,28 +30,32 @@ server.use(restify.gzipResponse());
...
 server.use(blpSession.getSession);
 server.use(requestHandler.elevateRequest);
 
-// Route
-server.post('/request/:ns/:svName/:reqName',
+// Routing
+
+// `/request` handles request/response services. There is one operation on these services, namely
+// "get me some data". In some sense this ought to be a GET but because the request data is fairly
+// large we use a POST.
```
ericvw: I would be OK if this was a `GET` instead of a `POST`.  `POST` is really intended for indicating that the state of the server is mutated, where `GET` does not imply so.  Since it is atypical for data for `GET` requests, it is not forbidden.

lib/blpapi-wrapper.ts
```diff
@@ -69,15 +69,29 @@ var EVENT_TYPE = {
   , PARTIAL_RESPONSE: 'PARTIAL_RESPONSE'
 };
 
+// Mapping of request types to response names to listen for.
+// The strings are taken from section A of the BLPAPI Developer's Guide, and are organized by
+// service.
+var REQUEST_TO_RESPONSE_MAP: { [index: string]: string; } = {
```
ericvw, line 75: Part of me wants to advocate for a switch/function for the fall-through for `fieldResponse`.  However, I don't feel that strongly about it.

lib/blpapi-wrapper.ts
```diff
@@ -272,15 +286,19 @@ export class Session extends events.EventEmitter {
         }).nodeify(cb);
     }
 
-    request(uri: string, name: string, request: any, callback: IRequestCallback): void {
+    request(uri: string, requestName: string, request: any, callback: IRequestCallback): void {
         this.validateSession();
 
+        var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
+        if (responseEventName === undefined) {
+            callback(new Error('Request type ' + requestName + ' not handled'));
```
ericvw: I would re-word the error message to indicate that it is a user error, not an internal error if the request name is incorrect.  Maybe something along the lines of `Unknown request type: ' + requestName`

lib/config.ts
```diff
@@ -242,8 +242,10 @@ export function get(name: string): any
     };
 
     // Restify bodyParser plugin options
-    otherConf['bodyParserOptions']
-        = { maxBodySize: convictConf.get('maxBodySize') };
+    otherConf['bodyParserOptions'] = {
+        maxBodySize: convictConf.get('maxBodySize'),
+        mapParams: false
```
ericvw, line 247: What are the implications of this?

lib/middleware/api-session.ts
```diff
@@ -12,28 +12,6 @@ import Interface = require('../interface');
 var SESSIONSTORE = new SessionStore(conf.get('expiration'));
 
 // PUBLIC FUNCTIONS
-export function createSession(req: Interface.IOurRequest,
```
ericvw, line 15: What's the motivation for removing this?  It seems to be conflating the work done in `handleSession`, which seems to be doing quite a bit for different types of operations.

lib/middleware/request-handler.ts
```diff
@@ -191,11 +191,23 @@ export function onRequest(req: Interface.IOurRequest,
...
+        // TODO: improve how we communicate errors (e.g. remove duplicates)
+        return next(new restify.BadRequestError('Invalid query params: ' + JSON.stringify(errors)));
+    }
+
     ((): Promise<any> => {
         return (new Promise<any>((resolve: () => void,
                                   reject: (error: any) => void): void => {
-            req.blpSession.request('//' + req.params.ns + '/' + req.params.svName,
-                req.params.reqName,
+            req.blpSession.request('//' + req.query.ns + '/' + req.query.service,
```
ericvw: I would recommend `util.format` from NodeJS.

lib/middleware/request-handler.ts
```diff
@@ -494,3 +494,36 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
+        // TODO: improve how we communicate errors (e.g. remove duplicates)
+        return next(new restify.BadRequestError('Invalid query params: ' + JSON.stringify(errors)));
+    }
+
+    switch (req.query.action) {
+        case 'subscribe':
+            return onSubscribe(req, res, next);
+        case 'unsubscribe':
+            return onUnsubscribe(req, res, next);
+    }
```
ericvw, line 493: Style and assert/error on default fall-through.

``` typescript
    switch (req.query.action) {
        case 'subscribe': {
            return onSubscribe(req, res, next);
        } break;
        case 'unsubscribe': { 
            return onUnsubscribe(req, res, next);
        } break;
        default: {
            assert(false, 'some descriptive message here');
        } break;
    }
```

lib/middleware/request-handler.ts
```diff
@@ -494,3 +494,36 @@ export function onUnsubscribe(req: Interface.IOurRequest,
             return next(new restify.InternalError(err.message));
         });
 }
+
+export function onChangeSubscriptions(req: Interface.IOurRequest,
+                                      res: Interface.IOurResponse,
+                                      next: restify.Next): void
+{
+    if (!req.apiSession) {
```
ericvw: Under what situation can this happen?

lib/middleware/request-handler.ts
```diff
@@ -494,3 +494,36 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
+        req.log.error('No apisession object found.');
+        return next(new restify.InternalError('No apisession object found.'));
+    }
+
+    if (!req.blpSession) {
```
ericvw: Under what situation can this happen?

lib/blpapi-wrapper.ts
```diff
@@ -272,15 +286,19 @@ export class Session extends events.EventEmitter {
         }).nodeify(cb);
     }
 
-    request(uri: string, name: string, request: any, callback: IRequestCallback): void {
+    request(uri: string, requestName: string, request: any, callback: IRequestCallback): void {
         this.validateSession();
 
+        var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
+        if (responseEventName === undefined) {
```
ericvw: Please change this test to be to protect against having an accidental mapping where the value is `undefined`:

``` typescript
if (!(responseEventName in REQUEST_TO_RESPONSE_MAP)) {
    return callback(new Error('Request type ' + requestName + ' not handled'));
}
var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
```

lib/config.ts
yupinghu, line 247: `mapParams` tells restify whether to dump the parse results into the `params` of the request in addition to `body` or `query` (or elsewhere, but right now we're only using those two). Setting it to false means we have to look specifically in the right place, rather than just checking `params` for everything, but also minimizes collisions.

lib/middleware/api-session.ts
yupinghu, line 15: When you have different routes for subscribe and unsubscribe, you could have different functions called for each. However, we now have just one route for `POST /realtime`, so we have just one function called by that route. If there's a way in restify to tell the difference earlier, I'll be happy to revert to the old setup, I just didn't think there was.

lib/middleware/request-handler.ts
yupinghu: Not sure, this was the existing code (moved to a new function) and may have just been defensive. @ericlu88 ?

index.ts
yupinghu, line 49: Yeah, that's why I left it, but now that I've added comments, I think this doesn't need to align with the following section, so I'll join 'em.

index.ts
yupinghu: Switched to GET, can always switch back if it's confusing for users.

lib/middleware/request-handler.ts
yupinghu, line 493: Created #75 for adding to the style guide as well.

index.ts
yupinghu: Never mind, switched back! See: http://stackoverflow.com/questions/978061/http-get-with-request-body

lib/middleware/request-handler.ts
ericlu88: This should probably be an assert, since this should never happen, unless something wrong with the `api-session` middleware.

lib/middleware/request-handler.ts
ericlu88: Yes, it is defensive. See my previous comments above.

lib/middleware/request-handler.ts
```diff
@@ -191,11 +192,23 @@ export function onRequest(req: Interface.IOurRequest,
         return next(new restify.InternalError('Error not find blpSession.'));
     }
 
+    // Validate the query parameters
```
ericlu88: I will advocate moving these checks into an util middleware

lib/middleware/request-handler.ts
```diff
@@ -323,16 +326,24 @@ export function onPoll(req: Interface.IOurRequest,
         return next(new restify.BadRequestError('No active subscriptions.'));
     }
 
+    req.checkQuery('pollid', 'Invalid pollid').isInt();
```
ericlu88: If we have a separate middleware to handle checking query parameters, we can configure it to check if `pollid` present and let the route handle validating the content.

yupinghu:
I rebased from bloomberg/develop to pick up more examples before fixing the examples, so that's why the whole "added some commits" thing happened.

typings/restify-validator/restify-validator.d.ts
```diff
@@ -0,0 +1,24 @@
+// Type definitions for adding validator to restify
```
ericlu88: I think we should put this typing file alongside with the `express-validator` since it is not a `restify-validator`.

typings/restify-validator/restify-validator.d.ts
```diff
@@ -0,0 +1,24 @@
...
+        checkParams(field:string, message:string): validator.Validator;
+        checkQuery(field:string, message:string): validator.Validator;
+        checkHeader(field:string, message:string): validator.Validator;
+        sanitize(field:string): validator.Sanitizer;
+        onValidationError(func:(msg:string) => void): void;
+        validationErrors(mapped?:boolean) : any;
+    }
+}
+
+declare module "express-validator" {
```
ericlu88: Is this section needed for the compiler?

lib/middleware/util.ts
```diff
@@ -67,3 +67,25 @@ export function rejectHTTP(req: Interface.IOurRequest,
...
 }
+
+// Uses express-validator to check the request for any problems. Callers pass a function that
+// performs whatever validation checks are necessary, and the returned RequestHandler will do the
+// boilerplate. The returned function is named and not just an anonymous function in order for
+// logging purposes.
+export function validateRequest(check: (req: Interface.IOurRequest) => void): restify.RequestHandler
+{
+    function validateRequest(req: Interface.IOurRequest,
+            res: Interface.IOurResponse,
```
ericvw: style: Align parameters

typings/restify-validator/restify-validator.d.ts
yupinghu: Yes, because validator() returns an express.RequestHandler and we need to instead treat it as a restify.RequestHandler.

typings/restify-validator/restify-validator.d.ts
yupinghu: The main thing is that I want it to be a separate file (so that we can grab updates to `express-validator.d.ts` without thinking too hard), but if what you're suggesting is to put it in the same directory, I'm ok with that. Suggestions for names? `express-to-restify-validator`? :)

index.ts
```diff
@@ -29,28 +31,43 @@ server.use(restify.gzipResponse());
...
 
-// Route
-server.post('/request/:ns/:svName/:reqName',
+// Routing
+
+// `/request` handles request/response services. There is one operation on these services, namely
+// "get me some data". In some sense this ought to be a GET but because the request data is large
+// we put it in the body.
+server.post('/request',
+            util.validateRequest((req: Interface.IOurRequest): void => {
```
ericlu88: I think instead of having the general `validateRequest` function exposed, we can have a small `util` module/class that expose helper functions for each route, so we don't have to stuff specific parameters inside `index.ts`

typings/restify-validator/restify-validator.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    import restify = require('restify');
+    function ExpressValidator(middlewareOptions?:any):restify.RequestHandler;
```
ericlu88: need a space after the colon.

typings/restify-validator/restify-validator.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/// <reference path="../express-validator/express-validator.d.ts" />
+/// <reference path="../restify/restify.d.ts" />
+
+declare module "restify" {
+    import http = require('http');
+    import validator = require('express-validator');
+
+    interface Request extends http.ServerRequest {
+        checkBody(field:string, message:string): validator.Validator;
```
ericlu88: need a space between the parameter name and the type.

typings/restify-validator/restify-validator.d.ts
```diff
@@ -0,0 +1,24 @@
...
+        checkParams(field:string, message:string): validator.Validator;
+        checkQuery(field:string, message:string): validator.Validator;
+        checkHeader(field:string, message:string): validator.Validator;
+        sanitize(field:string): validator.Sanitizer;
+        onValidationError(func:(msg:string) => void): void;
+        validationErrors(mapped?:boolean) : any;
```
ericlu88: need to remove a space before the last colon.

index.ts
yupinghu: I like having the param values in `index.ts` because then you can see in one place what all the query params ought to be, but I did create some wrappers to make this a lot clearer.

lib/blpapi-wrapper.ts
```diff
@@ -272,17 +286,19 @@ export class Session extends events.EventEmitter {
...
         var correlatorId = this.nextCorrelatorId();
         this.requests[correlatorId] = callback;
 
         this.openService(uri).then((): void => {
-            var responseEventName = reqName2RespName(name);
-            var requestName = name + 'Request';
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
+            if (!(requestName in REQUEST_TO_RESPONSE_MAP)) {
+                return callback(new Error('Unknown reques type ' + requestName));
```
ericvw: s/reques/request/

lib/middleware/request-handler.ts
```diff
@@ -494,3 +470,25 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
+
+    switch (req.query.action) {
+        case 'start': {
+            return onSubscribe(req, res, next);
+        } break;
+        case 'stop': {
+            return onUnsubscribe(req, res, next);
+        } break;
+        default: {
+            var errorString = 'Invalid realtime action: ' + req.query.action;
```
ericvw: s/realtime/subscription

ericlu88: Invalid subscription action.

ericvw:
lgtm - please squash and ensure that the commit has an appropriate summary line and body message that describes at a high-level the changes being landed.  Thanks!

index.ts
```diff
@@ -29,28 +32,35 @@ server.use(restify.gzipResponse());
...
+            requestValidator.requireQueryParams(['ns', 'service', 'type']),
             requestHandler.onRequest);
-server.post('/subscribe',
-            util.rejectHTTP,
-            apiSession.createSession,
-            apiSession.handleSession,
-            requestHandler.onSubscribe);
-server.post('/unsubscribe',
+
+// `/realtime` handles subscription services. There are two things you can do with subscriptions:
```
ericvw: s/realtime/subscription

yupinghu:
Amended yet again! :)

ericvw:
lgtm.  @ericlu88?

ericlu88:
LGTM. I will merge.

PR #74, opened 2015-02-13
Added instrument, curve, government lookups

thmiceli:
This is a continuation of #64 but bases on a branch forked out of bloomberg/blpapi-http.

I took into account the feedback provided so far. The request map has been simplified since it was decided that we will keep the actual request name in the URL (see discussion for #59).
I also added some additional examples.


thmiceli:
Just noticed that there is some overlap with what #73.

lib/blpapi-wrapper.ts
```diff
@@ -279,8 +286,11 @@ export class Session extends events.EventEmitter {
         this.requests[correlatorId] = callback;
 
         this.openService(uri).then((): void => {
-            var responseEventName = reqName2RespName(name);
-            var requestName = name + 'Request';
+            if (!(name in REQUEST_NAMES_MAP)) {
```
ericlu88, line 289: @yupinghu is making similar changes here in #73. Maybe we can land that first and rebase against that?

ericvw, line 289: +1

ericvw:
#73 landed.

thmiceli:
Superseeded by #83

Issue #57, opened 2015-02-09
Support //blp/tasvc


ericvw:
This issue is to track the progress of adding support for the `//blp/tasvc` service.  Please update with progress, findings, and status.

thmiceli:
Partially implemented with PR #83.
Still waiting for example for a subscription to technical analysis service.

thmiceli:
I haven't been able to get a working example for subscription to tasvc.
I may be able to see what's happening by debugging the C++ API calls.

ericvw:
I'm removing this from this milestone to reschedule for another.

Issue #55, opened 2015-02-09
Support //blp/mktbar


ericvw:
This issue is to track the progress of adding support for the `//blp/mktbar` service.  Please update with progress, findings, and status.

thmiceli:
We are currently extracting the subscription service name from the prefix of the security (defaulting to '//blp/mktdata' when not specified).
Do we want to support scenarios where within the same array of subscriptions, securities have different services specified? I would think so but I'd like make sure.

yupinghu:
A quick reading of the current subscription implementation suggests this already should work, but it's worth testing.

thmiceli:
@yupinghu yes it does. I tried mixing mktdata and mktvwap and it works. But we need to be able to listen to events other than just MarketDataEvents, I am working on that.

thmiceli:
Closed with PR #83

Issue #54, opened 2015-02-09
Support //blp/mktvwap


ericvw:
This issue is to track the progress of adding support for the `//blp/mktvwap` service.  Please update with progress, findings, and status.

thmiceli:
Closed with PR #83

PR #83, opened 2015-02-19
Additional services

thmiceli:
This changeset addresses the following issues: #54, #55, #55, #57 and #57.
I am still waiting for some answers for completing #57 (subscription to //blp/tasvc) though.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -84,7 +84,23 @@ var REQUEST_TO_RESPONSE_MAP: { [index: string]: string; } = {
     // //blp/apiflds
     'FieldInfoRequest':              'fieldResponse',
     'FieldSearchRequest':            'fieldResponse',
-    'CategorizedFieldSearchRequest': 'categorizedFieldResponse'
+    'CategorizedFieldSearchRequest': 'categorizedFieldResponse',
+
+    // //blp/instruments
+    'instrumentListRequest': 'InstrumentListResponse',
+    'curveListRequest':      'CurveListResponse',
+    'govtListRequest':       'GovtListResponse',
+
+    // //blp/tasvc
+    'studyRequest':          'studyResponse'
+};
+
+// Mapping of service URIs to the event names to listen to when subscribed to these services.
+var SERVICE_TO_SUBSCRIPTION_EVENTS_MAP: { [uri: string]: string[]; } = {
+    '//blp/mktdata':  ['MarketDataEvents'],
+    '//blp/mktvwap':  ['MarketDataEvents'],
+    '//blp/mktbar':   ['MarketBarStart', 'MarketBarUpdate', 'MarketBarEnd'],
+    '//blp/pagedata': ['PageUpdate']
 };
 
 // ANONYMOUS FUNCTIONS
@@ -92,14 +108,23 @@ function isObjectEmpty(obj: Object): boolean {
     return (0 === Object.getOwnPropertyNames(obj).length);
 }
 
-function subscriptionsToServices(subscriptions: Subscription[]): string[] {
-    return _(subscriptions).map((subscription: Subscription): string => {
-        var serviceRegex = /^\/\/blp\/[a-z]+/;
-        var match = serviceRegex.exec(subscription.security);
-        // XXX: note that we shoud probably capture what the default service is to use when
-        //      reading in the session options.  However, when not specified, it is
-        //      '//blp/mktdata'.
-        return match ? match[0] : '//blp/mktdata';
+function getServiceForSecurity(security: string): string {
+    var serviceRegex = /^\/\/blp\/[a-z]+/;
+    var match = serviceRegex.exec(security);
+    // XXX: note that we shoud probably capture what the default service is to use when
+    //      reading in the session options.  However, when not specified, it is
+    //      '//blp/mktdata'.
+    return match ? match[0] : '//blp/mktdata';
+}
+
+type SubscriptionAndService = {
+    subscription: Subscription;
+    serviceUri: string;
+}
+
+function getUniqueServices(sass: SubscriptionAndService[]): string[] {
+    return _(sass).map((subscriptionAndService: SubscriptionAndService): string => {
+        return subscriptionAndService.serviceUri;
     }).uniq().valueOf();
 }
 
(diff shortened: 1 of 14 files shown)


yupinghu:
Something's off with the history above. :) Has this been rebased off origin/develop recently?

lib/blpapi-wrapper.ts
```diff
@@ -99,7 +116,7 @@ function subscriptionsToServices(subscriptions: Subscription[]): string[] {
         // XXX: note that we shoud probably capture what the default service is to use when
         //      reading in the session options.  However, when not specified, it is
         //      '//blp/mktdata'.
-        return match ? match[0] : '//blp/mktdata';
+        return (subscription.serviceUri = match ? match[0] : '//blp/mktdata');
```
yupinghu: If not for seeing this in the diff, I would likely miss the assignment. Maybe break this into two statements for clarity.

lib/blpapi-wrapper.ts
```diff
@@ -336,9 +353,18 @@ export class Session extends events.EventEmitter {
...
-                this.listen('MarketDataEvents', cid, (m: any): void => {
-                    s.emit('data', m.data, s);
-                });
+
+                // Lookup by service the events we need to listen to.
+                if (!(s.serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP)) {
+                    throw new Error('Unknown service: ' + s.serviceUri);
+                }
+                var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[s.serviceUri];
+                for (var i: number = 0 ; i < events.length ; i++) {
```
yupinghu: Is there any style or other reason to prefer _.forEach over this form of for loop? @ericvw might have an opinion?

thmiceli:
I didn't do a rebase but a pull then a merge into my feature branch.

yupinghu:
Well, the main thing is that when we merge this PR, the commit history should get cleaned up first. This may be a side effect of the regrettable history rewriting @ericvw needed to do the other day? Either way, please see about fixing that history up.

lib/blpapi-wrapper.ts
thmiceli: No reason on my part other than it seemed overkill to set a callback for what needs to be done in the body of the loop or is there a forEach that doesn't take a callback?

ericvw:
Let's rebase against latest origin `develop` branch.  Otherwise this will be difficult to review.  My apologies for doing a force push :/.

lib/blpapi-wrapper.ts
ericvw: The JS engine is fairly efficient when doing `forEach`, functional style iteration.  The main advantage is it reduces the likelihood of side-effects and variable collisions because of function-level scope.  I tend to prefer `forEach` if I know I am iterating over every element and revert to for-loops if I plan on mucking with the indices.

thmiceli:
Pushed again after rebase.

yupinghu:
LGTM. Someone else should take a look as well, and then squash it down before submitting.

lib/blpapi-wrapper.ts
```diff
@@ -99,7 +116,8 @@ function subscriptionsToServices(subscriptions: Subscription[]): string[] {
         // XXX: note that we shoud probably capture what the default service is to use when
         //      reading in the session options.  However, when not specified, it is
         //      '//blp/mktdata'.
-        return match ? match[0] : '//blp/mktdata';
+        subscription.serviceUri = match ? match[0] : '//blp/mktdata';
```
ericvw: Should we modifying the subscription passed-in here?

lib/blpapi-wrapper.ts
```diff
@@ -336,8 +354,17 @@ export class Session extends events.EventEmitter {
...
                 //      until ES6 Map is available
                 this.subscriptions[cid] = s;
-                this.listen('MarketDataEvents', cid, (m: any): void => {
-                    s.emit('data', m.data, s);
+
+                // Lookup by service the events we need to listen to.
+                if (!(s.serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP)) {
+                    throw new Error('Unknown service: ' + s.serviceUri);
+                }
+                var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[s.serviceUri];
```
ericvw: I don't think there is a need to capture the `events` variable.  This could all be done functionally in-line.

lib/blpapi-wrapper.ts
ericvw: Yeah, it looks like it is.  Let's see if we can find a different approach that doesn't modify the passed-in subscriptions from the caller.

lib/blpapi-wrapper.ts
thmiceli: I found:

var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[s.serviceUri];
events.forEach((event: string): void => {...

more readable than:

SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[s.serviceUri].forEach((event: string): void => {...

lib/blpapi-wrapper.ts
thmiceli: Please see my latest update to the PR.

thmiceli:
Removed the service URI from the subscription class. Updated PR.

ericvw:
Now it lgtm.  @yupinghu, can I get a second pair of eyes on the latest changes?

lib/blpapi-wrapper.ts
```diff
@@ -321,23 +348,38 @@ export class Session extends events.EventEmitter {
             if (undefined !== cid) {
                 throw new Error('Subscription already exists for index ' + i);
             }
+
+        }).map((s: Subscription): SubscriptionAndService => {
+            return { subscription: s, serviceUri: getServiceForSecurity(s.security) };
```
ericvw: Style: Please conform to the style guide with respect to [object literals](http://github.com/bloomberg/blpapi-http/blob/6e27373b1b7219fe910ae5ab7879bab235c07aa9/doc/style-guide.md#object-and-array-literals).

yupinghu:
LGTM. Address the style issue above and squash. Thanks!

PR #85, opened 2015-02-19
handling of 'error' event for subscriptions

### LINKED ISSUE #39: SessionTerminated event is not received while socket is connected
When a socket is connected and I kill the underlying `blpSession`, I did not received a `SessionTerminated` events. The event is received if there's no socket connected.

ericvw:
NodeJS's `EventEmitter` treats emitting 'error' as a special case - when
there are no listeners, an `Error` object is thrown on behalf of the
`EventEmitter`.  Due to this behavior, an unhandled exception was being
thrown from `blpapi-wrapper.ts` that wasn't allowing the
`SessionTerminated` event being propagated properly to the users of the
API.

This change now has subscriptions listen for the 'error' event to
properly update the internal state of outstanding subscriptions.

Closes #39

### DIFF
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -265,6 +265,17 @@ export function onSubscribe(req: Interface.IOurRequest,
                 // Buffer the current data
                 sub.buffer.pushValue(data);
             });
+
+            // Must subscribe to the 'error' event; otherwise EventEmitter will throw an exception
+            // that was occurring from the underlying blpapi.Session.  It is the assumed that the
+            // blpapi.Session properly cleans up the subscription (i.e., 'unsubscribe()' should not
+            // be called).
+            sub.on('error', (err: Error): void => {
+                req.log.error(err, 'blpapi.Session subscription error occurred.');
+                sub.removeAllListeners();
+                req.apiSession.activeSubscriptions.delete(sub.correlationId);
+                req.apiSession.receivedSubscriptions.delete(sub.correlationId);
+            });
         });
 
         // Subscribe user request through blpapi-wrapper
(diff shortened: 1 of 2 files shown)


lib/websocket/websocket-handler.ts
```diff
@@ -121,10 +121,18 @@ function setup(socket: Interface.ISocket): void
...
-                                     'data': data});
+                socket.send('data', {
+                    'correlationId': sub.correlationId,
+                    'data': data
+                });
                 socket.log.info('Data sent');
             });
+
+            sub.on('error', (err: Error): void => {
+                sub.removeAllListeners();
```
yupinghu, line 138: The ordering differs between here and the other file. Intentional or unimportant?

ericvw, line 138: Unimportant, but I will fix it.

lib/middleware/request-handler.ts
```diff
@@ -265,6 +265,12 @@ export function onSubscribe(req: Interface.IOurRequest,
                 // Buffer the current data
                 sub.buffer.pushValue(data);
             });
+
```
ericlu88, line 268: It would be better to add a comments about why we need this event listener, and specifically pointing out the it doesn't need to **unsubscribe** when `error` occurs.

ericvw:
Updated PR based on latest comments from @yupinghu and @ericlu88.

ericvw:
PR updated based on out-of-band discussion with @ericlu88 about logging and sending an error back to the client.

ericlu88:
LGTM.

ericvw:
Squashed commits

PR #86, opened 2015-02-20
Simplify onResponse to remove unneeded promise layers.

yupinghu:
The wrapper's request function is already asynchronous, and
now that session handling is done earlier in middleware, we
can make requests a bit simpler. I also took the opportunity
to simplify how the response text is formed.

It should be possible to do the same sort of thing for
subscriptions but that will be a separate change.

### DIFF
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -119,66 +119,48 @@ export function elevateRequest (req: Interface.IOurRequest,
                                 next: restify.Next): void
 {
     var chunkIndex: number = 0;
-    var chunk: string; // the current chunk string is assembled here
-    var needComma: boolean = false; // need to append a ',' before adding more data
 
-    // returns a promise
-    function prepareResponse (): Promise<any> {
-        var resultP: Promise<any> = undefined;
-        if (++chunkIndex === 1) {
+    /**
+     * This is called each time data is added to the response.
+     * @param isEnd Whether this is the last time this will be called for this http request.
+     * @returns {string} Text that should be added to the response before the caller adds its data.
+     */
+    function prepareResponseText(isEnd: boolean): string {
+        var text: string;
+        if (1 === ++chunkIndex) {
+            // This is the first time we're adding response text. Set the headers as well.
+            res.statusCode = 200;
             res.setHeader('content-type', 'application/json');
-            chunk = '{';
-            needComma = false;
+            // If isEnd is true, then we aren't actually adding a data array to the http response.
+            text = isEnd ? '{' : '{"data":[';
         } else {
-            chunk = '';
+            // If isEnd is true, we must have previously added data to the response, so we need to
+            // close the data array.
+            text = isEnd ? '],' : ',';
         }
-
-        return resultP || Promise.resolve(undefined);
-    };
+        assert(text, 'Response text was not set');
+        return text;
+    }
 
     function stringifyPair ( key: string, value: any ): string {
         return '"' + key.replace(/'/g, '\\$&') + '":' + JSON.stringify(value);
     }
 
     res.sendChunk = (data: any): Promise<void> => {
-        var p = prepareResponse();
-        return p.then((): void => {
-            if (chunkIndex === 1) {
-                res.statusCode = 200;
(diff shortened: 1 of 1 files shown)


lib/middleware/request-handler.ts
```diff
@@ -190,30 +174,19 @@ export function onRequest(req: Interface.IOurRequest,
...
-                    }
-                });
-        }));
-    })()
-        .then(next)
-        .catch( (err: Error): any => {
-            req.log.error(err, 'Request error.');
-            return next(new restify.InternalError(err.message));
+    req.blpSession.request(util.format('//%s/%s', req.query.ns, req.query.service),
+        req.query.type,
```
ericvw: nit: Align parameters

lib/middleware/request-handler.ts
```diff
@@ -119,66 +119,50 @@ export function elevateRequest (req: Interface.IOurRequest,
...
-                chunk += ',';
-            }
-            chunk += stringifyPair( 'status', status ) + ',' +
-                     stringifyPair( 'message', message || '' ) +
-                     '}';
-            res.end( chunk );
-            needComma = false;
-            chunk = '';
-        });
+        var text: string = prepareResponseText(true);
```
ericvw: nit: Type declaration not needed - can use inference.

lib/middleware/request-handler.ts
```diff
@@ -119,66 +119,50 @@ export function elevateRequest (req: Interface.IOurRequest,
...
+                text = ',';
+            } else {
+                // This is the continuation of a response that has data and will not add more.
+                text = '],';
+            }
         }
-
-        return resultP || Promise.resolve(undefined);
-    };
+        return text;
```
ericvw, line 142: We should assert that `text` is set.

lib/middleware/request-handler.ts
```diff
@@ -119,66 +119,50 @@ export function elevateRequest (req: Interface.IOurRequest,
...
-    var chunk: string; // the current chunk string is assembled here
-    var needComma: boolean = false; // need to append a ',' before adding more data
 
-    // returns a promise
-    function prepareResponse (): Promise<any> {
-        var resultP: Promise<any> = undefined;
+    // This is called each time something
+    function prepareResponseText(isEnd: boolean): string {
+        var text: string;
         if (++chunkIndex === 1) {
```
ericvw: nit: Reverse operands for the equality

lib/middleware/request-handler.ts
```diff
@@ -119,66 +119,50 @@ export function elevateRequest (req: Interface.IOurRequest,
...
+            // This is the first time we're adding response text. Set the headers as well.
+            res.statusCode = 200;
             res.setHeader('content-type', 'application/json');
-            chunk = '{';
-            needComma = false;
+            if (!isEnd) {
```
ericlu88: How about using conditional operator `?` here for conciseness?

lib/middleware/request-handler.ts
```diff
@@ -119,66 +119,50 @@ export function elevateRequest (req: Interface.IOurRequest,
...
-        return resultP || Promise.resolve(undefined);
-    };
+        return text;
+    }
 
     function stringifyPair ( key: string, value: any ): string {
         return '"' + key.replace(/'/g, '\\$&') + '":' + JSON.stringify(value);
     }
 
     res.sendChunk = (data: any): Promise<void> => {
```
ericlu88, line 149: It seems to me that there's no async operation in `sendChunk` and `sendEnd` method anymore. We should probably make it completely synchronously(ie, `res.sendChunk = (data: any): void`)

yupinghu, line 149: Created #87 to track this -- I agree but subscriptions currently relies on this behavior and wasn't comfortable making that change all at once. I can tackle it after this one is in.

ericvw:
Note: Let's not merge this until #83 is landed first.

ericvw:
lgtm

ericlu88:
LGTM too. Will hold off merging until #83 landed.

ericvw:
@yupinghu - feel free to squash

PR #88, opened 2015-02-20
extend ISocket abstraction to hide message structure details

ericvw:
This change extends `interface ISocket` so that the message structure is
handled by a base class implementation.  `SocketBaseImpl` was introduced
to be a partial abstract-base-class that handles most of the `ISocket`
interface, while stubbing out with exceptions the methods that must be
implemented by the parent class.  The `Socket.IO` and `ws`
implementations are modified to subclass `SocketBaseImpl` and the APIs
called in `websocket-handler.ts` are updated appropriately.

### DIFF
--- a/lib/websocket/socket-base-impl.ts
+++ b/lib/websocket/socket-base-impl.ts
@@ -1,0 +1,88 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import events = require('events');
+import bunyan = require('bunyan');
+import blpapi = require('../blpapi-wrapper');
+import Interface = require('../interface');
+import emitterAdapter = require('./socket-event-emitter-adapter');
+
+export = SocketBaseImpl;
+
+class SocketBaseImpl extends emitterAdapter.SocketEventEmitterAdapter implements Interface.ISocket {
+
+    // DATA
+    private _logger: bunyan.Logger;
+    private _blpapiSession: blpapi.Session;
+
+    // PROTECTED MANIPULATORS
+    protected send(name: string, ...args: any[]): void {
+        throw new Error('send() must be implemented');
+    }
+
+    // CREATORS
+    constructor(logger: bunyan.Logger,
+                emitterToAdapt: emitterAdapter.EventEmitter = new events.EventEmitter())
+    {
+        super(emitterToAdapt);
+        this._logger = logger;
+    }
+
+    // MANIPULATORS
+    sendData(correlationId: number, data: any): void {
+        this.send('data', {
+            correlationId: correlationId,
+            data: data
+        });
+    }
+
+    sendError(message: string): void {
+        this.send('err', { message: message });
+    }
+
+    notifyConnected(): void {
+        this.send('connected');
+    }
+
+    notifySubscribed(): void {
+        this.send('subscribed');
+    }
+
+    notifyUnsubscribed(all: boolean = false): void {
+        var message = 'unsubscribe';
+        if (all) {
+            message += ' all';
+        }
+        this.send(message);
+    }
+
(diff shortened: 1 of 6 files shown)


lib/websocket/websocket-handler.ts
```diff
@@ -233,9 +240,7 @@ function setup(socket: Interface.ISocket): void
             activeSubscriptions.delete(s.correlationId);
             receivedSubscriptions.delete(s.correlationId);
         });
-        receivedSubscriptions.size
-            ? socket.send('unsubscribed')
-            : socket.send('unsubscribed all');
+        socket.notifyUnsubscribed(0 === receivedSubscriptions.size);
```
ericvw, line 243: I would appreciate a second look at this to ensure I did this right.

lib/websocket/socket-event-emitter-adapter.ts
```diff
@@ -52,9 +53,13 @@ class SocketEventEmitterAdapter implements events.EventEmitter {
     }
 
     emit(event: string, ...args: any[]): boolean {
-        // Note that we want to stub calls to 'emit' because SocketIO.Socket.emit is specialized
-        // and does not follow the the standard NodeJS.EventEmitter interface.
-        assert(false, '“emit” should not be called on SocketEventEmitterAdapter');
-        return false;
+        if (typeof this.emit === 'EventEmitter') {
```
ericlu88: `this.emitter === 'EventEmitter'`

lib/websocket/websocket-handler.ts
ericlu88, line 243: I guess we can simplify it in `socket.notifyUnsubscribe(!receivedSubscriptions.size)`

lib/websocket/socket-base-impl.ts
```diff
@@ -0,0 +1,89 @@
...
+    private _logger: bunyan.Logger;
+    private _blpapiSession: blpapi.Session;
+
+    // PROTECTED MANIPULATORS
+    protected send(name: string, ...args: any[]): void {
+        throw new Error('send() must be implemented');
+    }
+
+    // CREATORS
+    constructor(logger: bunyan.Logger, emitterToAdapt?: SocketIO.Socket) {
```
ericlu88: I prefer using default parameters for simplicity, so we can remove all the optional parameters and argument length check.

``` typescript
constructor(logger: bunyan.Logger, emitterToAdapt: SocketIO.Socket | events.EventEmitter = new events.EventEmitter()) {
```

lib/websocket/socket-event-emitter-adapter.ts
```diff
@@ -52,9 +52,13 @@ class SocketEventEmitterAdapter implements events.EventEmitter {
...
-        // Note that we want to stub calls to 'emit' because SocketIO.Socket.emit is specialized
-        // and does not follow the the standard NodeJS.EventEmitter interface.
-        assert(false, '“emit” should not be called on SocketEventEmitterAdapter');
-        return false;
+        if (this.emitter instanceof events.EventEmitter) {
+            return EVENTEMITTERPROTOTYPE.emit.apply(this.emitter, arguments);
+        } else {
+            // Note that we want to stub calls to 'emit' because SocketIO.Socket.emit is specialized
+            // and does not follow the the standard NodeJS.EventEmitter interface.
+            assert(false, '“emit” should not be called on SocketEventEmitterAdapter');
```
yupinghu: Nit: change the message to indicate SocketIO since emit works for some types of SEEAs.

yupinghu:
LGTM, squash!

ericvw:
squashed

PR #92, opened 2015-02-21
rely on .catch() to invoke callback with Error

ericvw:
This is necessary to make sure the callback is invoked asynchronously.


ericlu88:
LGTM.

PR #94, opened 2015-02-23
Add testing infrastructure and test cases for request/response.

ericlu88:
This PR adds testing infrastructure to facilitate writing end-to-end test cases for request/response.

For test cases, the routine is to create another `node` process that host the server(`TestHelper.ts` has some helper function for this), and let the test cases act as client to hit the server. When the test server boost up, it uses `mockery` to hijack the dependency to `blpapi-wrapper.ts` and substitute it with a mocking layer `mock-blpapi-wrapper.ts`, which behaves according to an environmental parameter set by the main process. See `test-request-response.ts` as a usable example.

This is the first bit of issue #18. The next flow will be adding subscription support in the test cases.


index.ts
```diff
@@ -90,6 +98,10 @@ if (conf.get('websocket.ws.enable')) {
     serverWS.listen(conf.get('websocket.ws.port'));
     serverWS.on('listening', (): void => {
         logger.info('ws server listening on', conf.get('websocket.ws.port') );
+        process.send({  // signal parent process that WS server is ready(for testing)
+            sType: 'WS',
+            message: 'server startup'
+        });
```
ericvw: As discussed, these separate `process.send` will be funnel-down into a single `Promise.all` call to be resolved and send back a signal to the parent process.

lib/test-infrastructure/mock-blpapi-wrapper.ts
```diff
@@ -0,0 +1,99 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+import events = require('events');
+import assert = require('assert');
```
ericvw: nit: sort imports

lib/test-infrastructure/mock-blpapi-wrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+            p = p.then((): void => {
+                    assert(this[s], 'Invalid operation ' + s);
+                    this[s].call(this, cb);
+                });
+        });
+    }
+
+    // Not yet implemented
+    subscribe(subscriptions: BAPI.Subscription[], cb?: (err: any) => void): Promise<void> {
+        throw new Error();
```
ericvw: Move `Not yet implemented` into the `Error()` as the message.

lib/test-infrastructure/mock-blpapi-wrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // Not yet implemented
+    unsubscribe(subscriptions: BAPI.Subscription[]): void {
+        throw new Error();
```
ericvw: Move `Not yet implemented` into the `Error()` as the message.

ericvw:
How do we feel about renaming `test-infrastructure` to just `test`?

lib/test-infrastructure/mock-blpapi-wrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+               Promise.reject(new Error('Can not stop blpsession'));
+    }
+
+    // Supported operations:
+    // sendPartialData, sendFinalData, sendError, terminateSession
+    request(uri: string, name: string, request: any, cb: BAPI.IRequestCallback): void {
+        assert(this.instructions.request, 'No request instructions.');
+        var p: Promise<void> = Promise.resolve();
+        _(this.instructions.request).forEach((s: string): void => {
+            p = p.then((): void => {
```
ericvw: Would using callbacks with either `setTimeout` or `process.nextTick` be more readable here?  I understand that Promises could be used for this, but a Promise typically returns a value as a result of some async operation.  It feels a bit awkward at first glance - still mulling it over though :).

lib/test-infrastructure/test-helper.ts
```diff
@@ -0,0 +1,47 @@
...
+    server.stdout.on('data', (data: any): void => {
+      console.log('stdout: ' + data);
+    });
+    server.stderr.on('data', (data: any): void => {
+      console.log('stderr: ' + data);
+    });
+    server.on('close', (data: any): void => {
+      console.log('child process exited with code ' + data);
+    });
+*/
```
ericvw: Can this comment be removed?

lib/test-infrastructure/test-index.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+// Setup mock blpapi-wrapper
+import assert = require('assert');
+import mockery = require('mockery');
+import MockWrapper = require('./mock-blpapi-wrapper');
+mockery.registerMock('../blpapi-wrapper', MockWrapper);
```
ericvw: nit: Add a line space between the imports and working code.

lib/test-infrastructure/mock-blpapi-wrapper.ts
ericlu88: As discussed, use `Promise.each()` for this functionality. Unfortunately, `.each()` is not exposed through the `bluebird` typing files. I've PR them for this change(borisyankov/DefinitelyTyped#3721)

lib/test-infrastructure/test-helper.ts
ericlu88: This is kind of useful for debugging purpose(ie, pipe the child process's stdout to the main process to figure out what's going on). Maybe add const to control whether to see it and set it if needed?

ericlu88:
`lib/test` is kind of confusing to me, would prefer some other names.

lib/test-infrastructure/test-helper.ts
ericvw: Let's use the `debug` module instead to control this.

ericvw:
Instead of `lib/test`, why don't we move the JS into the `test` folder instead.  I quick scan of other NodeJS projects do this as well.  We may need to restructure our tests to go within a subdirectory.

ericvw:
`should` just released 5.0.1.

ericlu88:
Updated `should` to 5.0.1

index.ts
```diff
@@ -93,4 +106,18 @@ if (conf.get('websocket.ws.enable')) {
     });
     var wss = new webSocket.Server({ server: serverWS.server });
     wss.on('connection', webSocketHandler.wsOnConnect);
+    pServers.push(new Promise<void>((resolve: () => void,
+                                     reject: (error: Error) => void): void => {
+        serverWS.on('listening', (): void => {
+            resolve();
+        });
+    }));
```
thmiceli: Very similar code is repeated 3 times with the only difference being the restify server variable.
I believe there is an opportunity here for refactoring into a function.

ericvw: +1

test/lib/mock-blpapi-wrapper.ts
```diff
@@ -0,0 +1,96 @@
...
+import Promise = require('bluebird');
+import _ = require('lodash');
+import BAPI = require('../../lib/blpapi-wrapper');
+
+export interface IInstruction {
+    start?: boolean;    // Default behavior is to start session successfully
+    stop?: boolean;     // Default behavior is to stop session successfully
+    request: string[];
+    //subscribe?: string[];
+    //unsubscribe?: string[];
```
thmiceli: Should the two comments above be removed?

yupinghu: I think an additional comment indicating they are not yet implemented might be more warranted. :)

yupinghu:
As we discussed, I think ultimately the mock wrapper and the real wrapper should be peers and share some code. It doesn't need to happen for this PR, but if deferred to a later change please do create an issue to make sure we don't forget about it.

Overall I think this PR makes sense. I'm still looking over the particulars to make sure I understand it and may have more comments.

ericlu88:
close this PR for now. Will create a new one since we decided to change path.

PR #96, opened 2015-02-23
Replace req.params with req.query in subscription

ericlu88:
Found one problem with the subscription code where we use `req.params` instead of `req.query`. Also update `long-polling` example to match the latest URL scheme.


yupinghu:
LGTM

PR #101, opened 2015-02-24
simplify error handling for mappings to underlying event messages

ericvw:
This change takes advantage of the
[blpapi-node](http://github.com/bloomberg/blpapi-node) throwing an
exception or emitting an error that will get handled naturally by the
top-level `catch` handler.

For request/response, an `assert` has been added to ensure that there is
no programmer error in keeping the request/response listener map
up-to-date.  Subscriptions does not need the check since the opening of
the service would fail beforehand; the dead code for subscriptions is
removed.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -320,12 +320,10 @@ export class Session extends events.EventEmitter {
         this.openService(uri).then((): void => {
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
-            if (!(requestName in REQUEST_TO_RESPONSE_MAP)) {
-                throw new Error('Unknown request type ' + requestName);
-            }
-            var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
             this.session.request(uri, requestName, request, correlatorId);
-            this.listen(responseEventName,
+            assert(requestName in REQUEST_TO_RESPONSE_MAP,
+                   util.format('Request, %s, not handled', requestName));
+            this.listen(REQUEST_TO_RESPONSE_MAP[requestName],
                         correlatorId,
                         this.requestHandler.bind(this, callback));
         }).catch((ex: Error): void => {
@@ -373,11 +371,9 @@ export class Session extends events.EventEmitter {
                 //      until ES6 Map is available
                 this.subscriptions[cid] = s;
 
-                // Lookup by service the events we need to listen to.
-                if (!(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP)) {
-                    throw new Error('Unknown service: ' + serviceUri);
-                }
-                var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[serviceUri];
+                assert(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP,
+                       util.format('Service, %s, not handled', serviceUri));
+                var events = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[serviceUri];
                 events.forEach((event: string): void => {
                     log('listening on event: ' + event + ' for cid: ' + cid);
                     this.listen(event, cid, (m: any): void => {


lib/blpapi-wrapper.ts
```diff
@@ -320,12 +320,10 @@ export class Session extends events.EventEmitter {
...
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
-            if (!(requestName in REQUEST_TO_RESPONSE_MAP)) {
-                throw new Error('Unknown request type ' + requestName);
-            }
-            var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
             this.session.request(uri, requestName, request, correlatorId);
-            this.listen(responseEventName,
+            assert(requestName in REQUEST_TO_RESPONSE_MAP,
+                   util.format('Request, %s, not handled'));
```
ericlu88: `util.format('Request %s not handled', requestName)`

lib/blpapi-wrapper.ts
```diff
@@ -373,11 +371,7 @@ export class Session extends events.EventEmitter {
                 //      until ES6 Map is available
                 this.subscriptions[cid] = s;
 
-                // Lookup by service the events we need to listen to.
-                if (!(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP)) {
-                    throw new Error('Unknown service: ' + serviceUri);
-                }
-                var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[serviceUri];
+                var events = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[serviceUri];
```
ericlu88, line 376: We need an `assert` here for the same reason for request/response, when the serviceUri is valid but we don't have a corresponding entry in the map.

ericlu88:
I guess we still needs a `catch` block for `subscribe` & `unsubscribe` method?

ericvw:
A `.catch` is not needed for `subscribe` nor `unsubscribe`.  For `subscribe`, the `.nodeify` call properly executes the callback with the thrown error or will be a rejected promise.  `unsubscribe` doesn't use a `Promise`.

ericvw:
Updated PR based on in-line feedback.

lib/blpapi-wrapper.ts
```diff
@@ -373,11 +371,9 @@ export class Session extends events.EventEmitter {
                 //      until ES6 Map is available
                 this.subscriptions[cid] = s;
 
-                // Lookup by service the events we need to listen to.
-                if (!(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP)) {
-                    throw new Error('Unknown service: ' + serviceUri);
-                }
-                var events: string[] = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[serviceUri];
+                assert(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP,
+                       util.format('Service, %s, not handled', serviceUri));
```
yupinghu, line 375: Nit: maybe 'Subscription service' (to mention subscriptions vs. request here)?

ericvw, line 375: This would get thrown during `subscribe()`; is that enough context?

yupinghu, line 375: Either way is fine by me, if the stack gets printed it'll be clear enough too. I generally like the message being super clear (C habit, I guess, where the stack trace isn't always that easy to read) but it'll be visible either way in this case.

lib/blpapi-wrapper.ts
```diff
@@ -320,12 +320,10 @@ export class Session extends events.EventEmitter {
...
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
-            if (!(requestName in REQUEST_TO_RESPONSE_MAP)) {
-                throw new Error('Unknown request type ' + requestName);
-            }
-            var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
             this.session.request(uri, requestName, request, correlatorId);
-            this.listen(responseEventName,
+            assert(requestName in REQUEST_TO_RESPONSE_MAP,
+                   util.format('Request, %s, not handled', requestName));
```
thmiceli, line 325: Isn't the requestName provided by the user? Is that the case we shouldn't assert if the lookup fails.

ericvw, line 325: `this.session.reqeust()` will already throw an exception and error appropriately.  So the `assert` is valid to ensure that our internal imp-details of the mapping is correct.

lib/blpapi-wrapper.ts
ericvw, line 375: We'll have an easy stack trace in this case.

thmiceli:
LGTM

ericvw:
@ericlu88 or @yupinghu, can I get a final look to get this merged?  I don't think there are ay other pending review items to address.

ericlu88:
LGTM.

yupinghu:
LGTM, squash and we'll merge.

ericvw:
Squashed

ericvw:
TravisCI is happy now.  Looks like it just hiccuped.

Issue #102, opened 2015-02-25
Have reasonable way(s) to get an auth token


yupinghu:
Related to #15.

Right now, we only have AuthorizationTokenRequest which sends you a token via MSG, which is lame.

PR #104, opened 2015-02-25
Basic authorization support.

### LINKED ISSUE #15: Support //blp/apiauth
This issues is to understand how the [Open API](http://www.bloomberglabs.com/api/) authorization service and system works.  Section 6 of the [BLP API Developers Guide](http://www.bloomberglabs.com/content/uploads/sites/2/2014/07/blpapi-developers-guide-2.54.pdf)  discusses the `//blp/apiauth` service, but we need a better understanding how this works in practice.

This investigation encompasses setting up our own BPIPE installation and user management to properly test authorization.

yupinghu:
This allows users to supply an authorization token and
tracks the created identities across requests (based on
client cert). Closes #15 in a narrow sense although the
feature is not very usable until #102 is also fixed.

Also includes a refactor of dispatching requests with action
query parameters.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -50,7 +50,7 @@ export class Subscription extends events.EventEmitter {
 
 export class BlpApiError implements Error {
     // STATIC DATA
-    static NAME: string = 'BlpApiErorr';
+    static NAME: string = 'BlpApiError';
 
     // DATA
     data: any;
@@ -59,7 +59,8 @@ export class BlpApiError implements Error {
     constructor(data: any) {
         this.data = data;
         this.name = BlpApiError.NAME;
-        this.message = data.reason.description;
+        // Subscription errors have a description, other errors have a message.
+        this.message = data.reason.message || data.reason.description;
     }
 }
 
@@ -92,7 +93,11 @@ var REQUEST_TO_RESPONSE_MAP: { [index: string]: string; } = {
     'govtListRequest':       'GovtListResponse',
 
     // //blp/tasvc
-    'studyRequest':          'studyResponse'
+    'studyRequest':          'studyResponse',
+
+    // //blp/apiauth
+    'AuthorizationRequest':      'AuthorizationResponse',
+    'AuthorizationTokenRequest': 'AuthorizationTokenResponse'
 };
 
 // Mapping of service URIs to the event names to listen to when subscribed to these services.
@@ -254,6 +259,37 @@ export class Session extends events.EventEmitter {
         log('Session terminated');
     }
 
+    private doRequest(uri: string,
+                      requestName: string,
+                      request: any,
+                      callback: IRequestCallback,
+                      isAuthRequest: boolean,
+                      identity?: blpapi.Identity): void
+    {
+        this.validateSession();
+
+        var correlatorId = this.nextCorrelatorId();
+        this.requests[correlatorId] = callback;
+
+        this.openService(uri).then((): void => {
+            log(util.format('Request: %s|%d', requestName, correlatorId));
+            trace(request);
+            if (isAuthRequest) {
+                this.session.authorizeUser(request, correlatorId);
+            } else {
+                this.session.request(uri, requestName, request, correlatorId, identity);
+            }
+            assert(requestName in REQUEST_TO_RESPONSE_MAP,
+                   util.format('Request, %s, not handled', requestName));
+            this.listen(REQUEST_TO_RESPONSE_MAP[requestName],
+                        correlatorId,
+                        this.requestHandler.bind(this, callback));
+        }).catch((ex: Error): void => {
+            delete this.requests[correlatorId];
+            callback(ex);
+        });
+    }
+
     // PRIVATE ACCESSORS
     private validateSession(): void {
         if (this.stopped) {
@@ -311,28 +347,47 @@ export class Session extends events.EventEmitter {
         }).nodeify(cb);
     }
 
-    request(uri: string, requestName: string, request: any, callback: IRequestCallback): void {
-        this.validateSession();
-
-        var correlatorId = this.nextCorrelatorId();
-        this.requests[correlatorId] = callback;
-
-        this.openService(uri).then((): void => {
-            log(util.format('Request: %s|%d', requestName, correlatorId));
-            trace(request);
-            this.session.request(uri, requestName, request, correlatorId);
-            assert(requestName in REQUEST_TO_RESPONSE_MAP,
-                   util.format('Request, %s, not handled', requestName));
-            this.listen(REQUEST_TO_RESPONSE_MAP[requestName],
-                        correlatorId,
-                        this.requestHandler.bind(this, callback));
-        }).catch((ex: Error): void => {
-            delete this.requests[correlatorId];
-            callback(ex);
-        });
+    request(uri: string,
+            requestName: string,
+            request: any,
+            identity: blpapi.Identity,
+            callback: IRequestCallback): void
+    {
+        var isAuthRequest = false;
+        this.doRequest(uri, requestName, request, callback, isAuthRequest, identity);
     }
 
-    subscribe(subscriptions: Subscription[], cb?: (err: any) => void): Promise<void> {
+    authorizeUser(request: any, cb?: (err: any, value: any) => void): Promise<blpapi.Identity> {
+        return new Promise<blpapi.Identity>((resolve: (identity: blpapi.Identity) => void,
+                                             reject: (err: Error) => void): void => {
+            var identity: blpapi.Identity = null;
+            function callback(err: Error, data?: any, isFinal?: boolean): void {
+                if (err) {
+                    reject(err);
+                } else {
+                    if (data.hasOwnProperty('identity')) {
+                        identity = data.identity;
+                    }
+                    if (isFinal) {
+                        if (identity) {
+                            resolve(identity);
+                        } else {
+                            reject(new BlpApiError(data));
+                        }
+                    }
+                }
+            }
+            var uri = '//blp/apiauth';
+            var requestName = 'AuthorizationRequest';
+            var isAuthRequest = true;
+            this.doRequest(uri, requestName, request, callback, isAuthRequest);
+        }).nodeify(cb);
+    }
+
+    subscribe(subscriptions: Subscription[],
+              identity: blpapi.Identity,
+              cb?: (err: any) => void): Promise<void>
+    {
         this.validateSession();
 
         var subscriptionsAndServices = _.forEach(subscriptions, (s: Subscription,
@@ -361,14 +416,12 @@ export class Session extends events.EventEmitter {
         })).then((): void => {
             log('Subscribing to: ' + JSON.stringify(subscriptions));
 
-            this.session.subscribe(_.map(subscriptionsAndServices,
-                                         (sas: SubscriptionAndService): blpapi.Subscription =>
-            {
+            function makeSubscription(sas: SubscriptionAndService): blpapi.Subscription {
                 var cid = this.nextCorrelatorId();
                 var s = sas.subscription;
                 var serviceUri = sas.serviceUri;
-                // XXX: yes, this is a side-effect of map, but it is needed for performance reasons
-                //      until ES6 Map is available
+                // XXX: yes, this is a side-effect of map, but it is needed for performance
+                //      reasons until ES6 Map is available
                 this.subscriptions[cid] = s;
 
                 assert(serviceUri in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP,
@@ -391,7 +444,10 @@ export class Session extends events.EventEmitter {
                     result.options = s.options;
                 }
                 return result;
-            }));
+            }
+
+            var subs = _.map(subscriptionsAndServices, makeSubscription);
+            this.session.subscribe(subs, identity);
         }).nodeify(cb);
     }
 
--- a/lib/interface.ts
+++ b/lib/interface.ts
@@ -1,5 +1,6 @@
 /// <reference path='../typings/tsd.d.ts' />
 
+import blpapi = require('blpapi');
 import restify = require('restify');
 import bunyan = require('bunyan');
 import BAPI = require('./blpapi-wrapper');
@@ -9,6 +10,7 @@ export interface IOurRequest extends restify.Request {
     clientCert?: any;
     blpSession: BAPI.Session;
     apiSession?: Session;
+    identity?: blpapi.Identity;
 }
 
 export interface IBufferedData<T> {
--- a/lib/middleware/auth.ts
+++ b/lib/middleware/auth.ts
@@ -1,0 +1,33 @@
+/// <reference path='../../typings/tsd.d.ts' />
+
+// Handles client auth operations.
+// Currently we only have authorization (e.g. Identity management), associating the identities with
+// client certs.
+
+import blpapi = require('blpapi');
+import _ = require('lodash');
+import restify = require('restify');
+import Interface = require('../interface');
+import Map = require('../util/map');
+
+var sIdentityMap: Map<blpapi.Identity> = new Map<blpapi.Identity>();
+
+export function getIdentity(req: Interface.IOurRequest,
+                            res: Interface.IOurResponse,
+                            next: restify.Next): void
+{
+    if (_.has(req, 'clientCert')) {
+        req.identity = sIdentityMap.get(req.clientCert.fingerprint) || null;
+    } else {
+        req.identity = null;
+    }
+    return next();
+}
+
+export function setIdentity(req: Interface.IOurRequest, identity: blpapi.Identity): void
+{
+    if (!_.has(req, 'clientCert')) {
+        throw new Error('Cannot setIdentity without a client cert');
+    }
+    sIdentityMap.set(req.clientCert.fingerprint, identity);
+}
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -1,6 +1,7 @@
 /// <reference path='../../typings/tsd.d.ts' />
 
 import assert = require('assert');
+import blpapi = require('blpapi');
 import Promise = require('bluebird');
 import _ = require('lodash');
 import restify = require('restify');
@@ -11,6 +12,26 @@ import conf = require('../config');
 import Subscription = require('../subscription/subscription');
 import Session = require('../apisession/session');
 import Map = require('../util/map');
+import BAPI = require('../blpapi-wrapper');
+import auth = require('./auth');
+
+// Type used for request handlers dispatching, i.e. where the top level request includes some
+// query param that we switch on to find the correct handler.
+type ActionMap = {
+    [index: string]: (req: Interface.IOurRequest,
+                      res: Interface.IOurResponse,
+                      next: restify.Next) => any;
+};
+
+var SUBSCRIPTION_ACTION_MAP: ActionMap = {
+    'start': onSubscribe,
+    'stop':  onUnsubscribe
+};
+
+var AUTH_ACTION_MAP: ActionMap = {
+    'token':     onGetToken,
+    'authorize': onAuthorize
+};
 
 // PRIVATE FUNCTIONS
 function validatePollId(session: Session, newPollId: number): { isValid: boolean;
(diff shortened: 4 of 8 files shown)


index.ts
```diff
@@ -53,7 +55,7 @@ server.post('/request',
 // and GET respectively.
 server.post('/subscription',
             util.rejectHTTP,
-            requestValidator.requireChoiceQueryParam('action', ['start', 'stop']),
+            requestValidator.requireActionQueryParam(requestHandler.getAuthActions()),
```
ericlu88: `requestValidator.requireActionQueryParam(requestHandler.getSubscriptionActions()),`

lib/interface.ts
```diff
@@ -21,6 +23,10 @@ export interface IOurResponse extends restify.Response {
     sendEnd?: (status: any, message: string) => void;
 }
 
+export interface IOurRequestHandler {
```
ericlu88: I don't see a specific need for this interface here. The only usage is in `request-handler.ts` as type alias , which I think we can put this declaration there.

lib/middleware/auth.ts
```diff
@@ -0,0 +1,35 @@
...
+                            res: Interface.IOurResponse,
+                            next: restify.Next): void
+{
+    if (_.has(req, 'clientCert')) {
+        var identity = identityMap.get(req.clientCert.fingerprint);
+        if (identity) {
+            req.identity = identity;
+        }
+    }
+    next();
```
ericlu88: For consistency, `return next()`

lib/blpapi-wrapper.ts
```diff
@@ -324,7 +334,11 @@ export class Session extends events.EventEmitter {
                 throw new Error('Unknown request type ' + requestName);
             }
             var responseEventName = REQUEST_TO_RESPONSE_MAP[requestName];
-            this.session.request(uri, requestName, request, correlatorId);
+            if (isAuthRequest) {
```
ericlu88: I am not a big fan of this code-reuse. I think I prefer having `request` & `authorizeUser` as separate implemented functions, cause I don't see lots of common behavior except the `listen` to event.

lib/blpapi-wrapper.ts
```diff
@@ -334,7 +348,25 @@ export class Session extends events.EventEmitter {
...
+            identity: blpapi.Identity,
+            callback: IRequestCallback): void
+    {
+        this.doRequest(uri, requestName, request, callback, false, identity);
+    }
+
+    authorizeUser(request: any, callback: IRequestCallback): void {
+        var uri = '//blp/apiauth';
+        var requestName = 'AuthorizationRequest';
+        this.doRequest(uri, requestName, request, callback, true, undefined);
```
ericlu88: Will the `AuthorizationResponse` be partial response? If not, I guess we can simplify the code here too. One way I am thinking is having `authorizeUser` returns `Promise<blpapi.Identify>`, and let the `request-handler` deal with setting it.

index.ts
```diff
@@ -62,6 +64,12 @@ server.get('/subscription',
            apiSession.handleSession,
            requestHandler.onPollSubscriptions);
 
+// `/auth` handles authorization related requests (and in the future, authentication).
+server.post('/auth',
+            util.rejectHTTP,
+            requestValidator.requireActionQueryParam(requestHandler.getAuthActions()),
+            requestHandler.onAuth);
```
ericvw: nit: argument alignment

typings/blpapi/blpapi.d.ts
```diff
@@ -19,19 +19,30 @@ declare module "blpapi" {
         correlation: number;
     }
 
+    // Opaque object representing an authorized Identity.
+    export interface Identity {
```
ericvw: This should be in the `blpapi.d.ts` and re-exported here.

lib/blpapi-wrapper.ts
yupinghu: Well, if this was two separate functions, each would be 20 lines, with exactly one line difference (`openService` promise + listening logic). I feel like this sharing is not great but it's better for future proofing to any other changes we might make here.

lib/blpapi-wrapper.ts
yupinghu: Well, even if it were multiple chunks, we could make a change of this sort because there's only 2 possible outcomes we care about communicating (the `Identity` for success, and an error message for failure). For now I was just maintaining identical behavior (i.e. as much as possible authorization doesn't look different from other requests), but am open to switching this up if we feel that's best.

lib/blpapi-wrapper.ts
ericvw: Have `doRequest` take a callback that executes the desired functionality?

lib/blpapi-wrapper.ts
```diff
@@ -395,7 +427,7 @@ export class Session extends events.EventEmitter {
                     result.options = s.options;
                 }
                 return result;
-            }));
+            }), identity);
```
ericvw: For clarity, it may be best to capture the `_.map` for subscribe in a variable so that the call to `subscribe` appears straightforward.

lib/middleware/auth.ts
```diff
@@ -0,0 +1,35 @@
...
+// Currently we only have authorization (e.g. Identity management), associating the identities with
+// client certs.
+
+import blpapi = require('blpapi');
+import _ = require('lodash');
+import restify = require('restify');
+import Interface = require('../interface');
+import Map = require('../util/map');
+
+var identityMap: Map<blpapi.Identity> = new Map<blpapi.Identity>();
```
ericvw: s/identityMap/IDENTITIES ?

lib/blpapi-wrapper.ts
thmiceli: As a general stylistic rule could we add a comment for each non explicit literal that is passed as argument to a function call? For example here:
"this.doRequest(uri, requestName, request, callback, /_isAuthRequest_/ true, /_blpapi.Identity_/ undefined);"

yupinghu: In this case I've captured them values in temp variables with useful names, as a form of documenting this.

lib/blpapi-wrapper.ts
yupinghu: I had callbacks before, and since there's exactly two types of requests in the underlying API, I switched to this form. I think this needs to be a shared function because of the difference is literally in the request type but am open to switching back to the callback form if you think that's really cleaner.

lib/middleware/auth.ts
yupinghu: How about `s_identityMap` to be consistent with the C++ naming convention for static variables?

yupinghu: ... or, because tslint doesn't like underscores, `sIdentityMap`?

lib/blpapi-wrapper.ts
```diff
@@ -311,7 +316,13 @@ export class Session extends events.EventEmitter {
         }).nodeify(cb);
     }
 
-    request(uri: string, requestName: string, request: any, callback: IRequestCallback): void {
+    private doRequest(uri: string,
```
ericvw: Please move `private` manipulators before `// CREATORS`

lib/blpapi-wrapper.ts
```diff
@@ -332,7 +347,49 @@ export class Session extends events.EventEmitter {
...
+        this.doRequest(uri, requestName, request, callback, isAuthRequest, identity);
+    }
+
+    authorizeUser(request: any, cb?: (err: any, value: any) => void): Promise<blpapi.Identity> {
+        return new Promise<blpapi.Identity>((resolve: Function, reject: Function): void => {
+            var finalData: any;
+            var identity: blpapi.Identity;
+            function callback(err: Error, data?: any, isFinal?: boolean): void {
+                if (err) {
+                    request.log.error(err, 'Authorization error');
```
ericvw: I don't think `.log` exists on this object :)

lib/blpapi-wrapper.ts
```diff
@@ -332,7 +347,49 @@ export class Session extends events.EventEmitter {
...
+            callback: IRequestCallback): void
+    {
+        var isAuthRequest = false;
+        this.doRequest(uri, requestName, request, callback, isAuthRequest, identity);
+    }
+
+    authorizeUser(request: any, cb?: (err: any, value: any) => void): Promise<blpapi.Identity> {
+        return new Promise<blpapi.Identity>((resolve: Function, reject: Function): void => {
+            var finalData: any;
+            var identity: blpapi.Identity;
```
ericvw: nit: init to `null`

lib/blpapi-wrapper.ts
```diff
@@ -332,7 +347,49 @@ export class Session extends events.EventEmitter {
...
+            function callback(err: Error, data?: any, isFinal?: boolean): void {
+                if (err) {
+                    request.log.error(err, 'Authorization error');
+                    reject(err);
+                } else {
+                    if (data.hasOwnProperty('identity')) {
+                        identity = data.identity;
+                    }
+                    if (isFinal) {
+                        finalData = data;
```
ericvw: When is `finalData` used?

lib/blpapi-wrapper.ts
```diff
@@ -332,7 +347,49 @@ export class Session extends events.EventEmitter {
...
+            request: any,
+            identity: blpapi.Identity,
+            callback: IRequestCallback): void
+    {
+        var isAuthRequest = false;
+        this.doRequest(uri, requestName, request, callback, isAuthRequest, identity);
+    }
+
+    authorizeUser(request: any, cb?: (err: any, value: any) => void): Promise<blpapi.Identity> {
+        return new Promise<blpapi.Identity>((resolve: Function, reject: Function): void => {
```
ericvw: nit: Please specify precise types here.

lib/middleware/auth.ts
```diff
@@ -0,0 +1,35 @@
...
+var sIdentityMap: Map<blpapi.Identity> = new Map<blpapi.Identity>();
+
+export function getIdentity(req: Interface.IOurRequest,
+                            res: Interface.IOurResponse,
+                            next: restify.Next): void
+{
+    if (_.has(req, 'clientCert')) {
+        var identity = sIdentityMap.get(req.clientCert.fingerprint);
+        if (identity) {
+            req.identity = identity;
```
ericlu88: I think it is equivalent to say `req.identity = sIdentityMap.get(req.clientCert.fingerprint);` here. If we want to differentiate when there's no identify object, we should set a `null` here explicitly.

lib/middleware/auth.ts
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+    return next();
+}
+
+export function setIdentity(req: Interface.IOurRequest, identity: blpapi.Identity): string
+{
+    if (!_.has(req, 'clientCert')) {
```
ericlu88, line 29: I think we could just `throw` here and let the user code handle the catch(which the `request-handler` code already does).

lib/middleware/request-handler.ts
```diff
@@ -455,24 +493,72 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
 export function onChangeSubscriptions(req: Interface.IOurRequest,
                                       res: Interface.IOurResponse,
                                       next: restify.Next): void
 {
-    assert(req.apiSession, 'apiSession not found');
-    assert(req.blpSession, 'blpSession not found');
+    dispatchAction(SUBSCRIPTION_ACTION_MAP, req, res, next);
+}
+
+function onGetToken(req: Interface.IOurRequest,
```
ericlu88: Is there any specific reason why we need the get token route for now? It seems like the same as an normal request to me.

lib/middleware/request-handler.ts
```diff
@@ -455,24 +493,72 @@ export function onUnsubscribe(req: Interface.IOurRequest,
...
-        } break;
-        case 'stop': {
-            return onUnsubscribe(req, res, next);
-        } break;
-        default: {
-            var errorString = 'Invalid subscription action: ' + req.query.action;
-            req.log.error(errorString);
-            return next(new restify.BadRequestError(errorString));
-        } break;
+function onAuthorize(req: Interface.IOurRequest,
```
ericlu88: nit: reorder the public/private functions in this module.

lib/middleware/request-handler.ts
yupinghu: Only to group it as part of `/auth` rather than as part of `/request`.

lib/middleware/auth.ts
yupinghu: It's not actually equivalent -- the current code means that `req` will have no property named `identity`, vs. having a property that is set to `null`. That said, I think explicitly setting it to `null` is more consistent with the discussion we've been having about this sort of thing.

yupinghu:
Updated, please take another look.

ericvw:
@yph, can you rebase and push.  TravisCI is not picking this PR up to build and there are active merge conflicts atm.  Thanks!

lib/websocket/websocket-handler.ts
```diff
@@ -145,7 +145,8 @@ function setup(socket: Interface.ISocket): void
         });
 
         // Subscribe user request through blpapi-wrapper
-        socket.blpSession.subscribe(subscriptions)
+        // TODO: Support authorized identity.
+        socket.blpSession.subscribe(subscriptions, undefined)
```
ericvw: s/undefined/null or make `subscribe()` take an optional identity object.

lib/middleware/request-handler.ts
```diff
@@ -390,92 +529,36 @@ export function onPollSubscriptions(req: Interface.IOurRequest,
...
-            if (!sub.buffer.isEmpty()) {
-                result.push(sub.buffer.startNewBuffer());
-            }
-            req.apiSession.activeSubscriptions.delete(sub.correlationId);
-            req.apiSession.receivedSubscriptions.delete(sub.correlationId);
-        });
-        req.log.debug({ activeSubscriptions:
-                       req.apiSession.activeSubscriptions.size },
-                      'Unsubscribed.');
+var SUBSCRIPTION_ACTION_MAP: ActionMap = {
```
ericvw: Let's put global constants at the top of the file between imports and functions.

lib/middleware/request-handler.ts
```diff
@@ -390,92 +529,36 @@ export function onPollSubscriptions(req: Interface.IOurRequest,
...
-        case 'stop': {
-            return onUnsubscribe(req, res, next);
-        } break;
-        default: {
-            var errorString = 'Invalid subscription action: ' + req.query.action;
-            req.log.error(errorString);
-            return next(new restify.BadRequestError(errorString));
-        } break;
-    }
+var AUTH_ACTION_MAP: ActionMap = {
```
ericvw: Let's put global constants at the top of the file between imports and functions.

lib/middleware/request-handler.ts
```diff
@@ -290,6 +239,196 @@ export function onSubscribe(req: Interface.IOurRequest,
...
+
+export function onRequest(req: Interface.IOurRequest,
+                          res: Interface.IOurResponse,
+                          next: restify.Next): void
+{
+    assert(req.blpSession, 'blpSession not found');
+
+    var uri = util.format('//%s/%s', req.query.ns, req.query.service);
+    var callback = getRequestCallback(req, res, next);
+    req.blpSession.request(uri, req.query.type, req.body, req.identity, callback);
```
ericvw: I understand what is happening here, but it seems misleading having this function take `next` without having it visible that it gets called.  I would advocate for calling next explicitly here or reworking the function signature and how everything fits together.

lib/middleware/request-handler.ts
```diff
@@ -232,10 +181,10 @@ export function onSubscribe(req: Interface.IOurRequest,
             }
 
             var sub = new Subscription(s.correlationId,
-                                       s.security,
-                                       s.fields,
-                                       s.options,
-                                       conf.get('longpoll.maxbuffersize'));
+                s.security,
```
ericvw: Did `tslint` not pick this up?

lib/middleware/request-handler.ts
```diff
@@ -212,9 +171,9 @@ export function onSubscribe(req: Interface.IOurRequest,
         }
 
         req.body.forEach((s: {'correlationId': number;
-                              'security': string;
-                              'fields': string[];
-                              'options'?: any }): void => {
+            'security': string;
```
ericvw: nit: indent

lib/middleware/request-handler.ts
```diff
@@ -290,6 +249,196 @@ export function onSubscribe(req: Interface.IOurRequest,
...
+}
+
+function onGetToken(req: Interface.IOurRequest,
+                    res: Interface.IOurResponse,
+                    next: restify.Next): void
+{
+    var callback = getRequestCallback(req, res, next);
+    var uri = '//blp/apiauth';
+    var requestName = 'AuthorizationTokenRequest';
+    req.blpSession.request(uri, requestName, req.body, undefined /* identity */, callback);
```
ericvw: I understand what is happening here, but it seems misleading having this function take `next` without having it visible that it gets called. I would advocate for calling next explicitly here or reworking the function signature and how everything fits together.

lib/middleware/request-handler.ts
```diff
@@ -11,6 +12,24 @@ import conf = require('../config');
 import Subscription = require('../subscription/subscription');
 import Session = require('../apisession/session');
 import Map = require('../util/map');
+import BAPI = require('../blpapi-wrapper');
+import auth = require('./auth');
+
+// Type used for request handlers dispatching, i.e. where the top level request includes some
+// query param that we switch on to find the correct handler.
+type ActionMap = { [index: string]: (req: Interface.IOurRequest,
```
ericvw: nit: Object-literal style guide

lib/middleware/request-handler.ts
ericvw: Actually, maybe a type alias may be more appropriate.  I meant the indent of `correlationId` on a new line to be in conformance with the object literal style guide, but that seems awkward here the more I think about.

lib/middleware/request-handler.ts
yupinghu: `next` is used in the callback, and inherently will never be called during the execution of `onRequest`. I can eliminate the `getRequestCallback` function to make the use of `next` obvious here, but that means either that I have to duplicate the callback code in `onGetToken` or have `onGetToken` call this `onRequest` in which case the call to `next` will not be explicit there.

More broadly, I don't think we can satisfy your concern here and in general refactor code that uses `next` -- if multiple route handlers want to perform the same logic using `next`, how would you handle that?

lib/middleware/request-handler.ts
yupinghu: Nope. I'll fix it but we should confirm the lint rule is correct.

ericvw: @thmiceli is on it

lib/middleware/request-handler.ts
yupinghu: For this change I'll preserve the indentation that existed (i.e. this code got reformatted due to my editor not preserving the indent when I moved this function). If we want to introduce a type alias we can do that separately.

ericvw:
lgtm - will defer to @ericlu88 for final pass

lib/middleware/request-handler.ts
```diff
@@ -99,109 +120,48 @@ function getAllOldBuffers(subscriptionStore: Map<Subscription>): Object[]
...
+                        next: restify.Next): void
 {
-    // Check the content type of the request
-    // TODO: configure acceptable content-type
-    if (!req.is('application/json')) {
-        req.log.debug('Unsupported Content-Type.');
-        return next(new restify.UnsupportedMediaTypeError('Unsupported Content-Type.'));
+    var action = req.query.action;
+    if (!(action in actionMap)) {
+        var errorString = util.format('Invalid action: ' + action);
```
ericlu88: minor: `util.format('Invalid action: %s', action)`

lib/websocket/websocket-handler.ts
```diff
@@ -145,7 +145,8 @@ function setup(socket: Interface.ISocket): void
         });
 
         // Subscribe user request through blpapi-wrapper
-        socket.blpSession.subscribe(subscriptions)
+        // TODO: Support authorized identity.
+        socket.blpSession.subscribe(subscriptions, null)
```
ericlu88, line 149: Created issue #119 for auth over websocket.

ericlu88:
LGTM. Please squash down and I will merge.

yupinghu:
squashed

Issue #98, opened 2015-02-24
Decide a consistent approach for error handling


ericlu88:
Right now in our request/response code we use `http` status code to negotiate the errors with client. This approach won't work for cases where we encounter errors in the middle of the request, since we are sending the data received from underlying session right away(not buffering).

The two possible approaches are:
- Buffer the response until the last bit and then send the data in 1 trunk, so that we could leverage the http status code.
- Always send **200** as http status code and leverage the `status` field in the response body to indicate the status of the response.

Open to discussion, the goal is to come up with a consistency approach for our server.

yupinghu:
Buffering is bad, possibly unacceptable as a general approach since we can have extremely large response data. So at a high level, my feeling is:
- If it's an error we can tell before we even start getting data from the API, generate a 4xx response (or 5xx if appropriate but I hope that'd be rare).
- If the very first chunk from the API indicates an error, we should consider whether we can make a 4xx/5xx here (but it's ok to also use the next strategy).
- If we get any valid data back from the API, then always return a 2xx. If we get an error in a later chunk, try to return as much valid data as possible and then set a property of the response (status, probably) to indicate what went wrong.

ericlu88:
For the second bullet point, I will prefer using the 3rd strategy of sending 2xx error for consistency purpose(the `request-handler` code doesn't need to differentiate between the first chunk or the following).

ericlu88:
Change my thought on the second point. Will let `res.sendError` figure out whether we are at the first chunk or not. Doing this will be more beneficial when we land #79 so we could infer the http code out of the error type.

Issue #79, opened 2015-02-18
blpapi-wrapper should return different error types


yupinghu:
Currently, any time the wrapper returns an error, we reject with that error, which gets handled as a 500. We should allow the wrapper to return various error types and surface that to the client.

ericvw:
@yupinghu, can you provide a scenario where we want to do this?

yupinghu:
As discussed in person: e.g. we want to be able to distinguish between bad inputs (e.g. invalid service specified) from, say, server errors, so that we can do a 4xx vs. a 5xx. Eventually things like auth errors will also need this to be distinguishable.

ericvw:
@thmiceli, is going to pick this up.

ericvw:
What's the status of this?

thmiceli:
I haven't actually looked into this yet.
Is there more to do than what was done in https://github.com/bloomberg/blpapi-node/pull/40

ericvw:
Just depending on the `blpapi-node` changes as a npm dependency update should be good enough.  This addresses the `blpapi-wrapper` emitting different error types.

thmiceli:
The dependency is already in package.json. So I don't think there is more to do for this issue. Right?

ericvw:
The dependency is in `package.json`, but we need a new release version of `blpapi-node` to depend upon that has the changes from bloomberg/blpapi-node@e863218ca407c62eb9d7360026c188e29cbf0612.

PR #106, opened 2015-02-25
Add sendError to response object

ericlu88:
This PR adds a `sendError` function which sent error message over the wire. We use this function to indicate error for cases we can't predetermined and need to go through the `blpapi`. This PR address #98.

### DIFF
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -146,6 +146,12 @@ export function elevateRequest (req: Interface.IOurRequest,
         return '"' + key.replace(/'/g, '\\$&') + '":' + JSON.stringify(value);
     }
 
+    // Right now we only use restify.InternalError(500)
+    // TODO: Add error type differentiation
+    function createError(err: Error): any {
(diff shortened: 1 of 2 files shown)


lib/middleware/request-handler.ts
```diff
@@ -159,6 +159,14 @@ export function elevateRequest (req: Interface.IOurRequest,
         res.end(text);
     };
 
+    res.sendError = (err: Error): void => {
+        var text = util.format('%s%s,%s}',
```
yupinghu: Should the body just be `sendEnd(-1, err.message)`?

ericlu88:
Change my thought on the second point in #79. `res.sendError` figure out whether we are at the first chunk or not. Doing this will be more beneficial when we land #79 so we could infer the http code out of the error type.

ericvw:
lgtm

yupinghu:
LGTM, squash it down!

thmiceli:
LGTM as far as I can tell.

ericlu88:
Squashed and rebased.

PR #109, opened 2015-02-27
Fixed source code formatting.

thmiceli:
Ran tslint with some custom enhancements.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -1,5 +1,9 @@
 {
     "rules": {
+        "align": [true,
+            "parameters",
+            "arguments",
+            "statements"],
         "class-name": true,
         "comment-format": {
             "check-space":     true,
@@ -35,6 +39,7 @@
         "quotemark": [true, "single"],
         "radix": true,
         "semicolon": true,
+        "switch-default": true,
         "triple-equals": true,
         "typedef": [true,
             "call-signature",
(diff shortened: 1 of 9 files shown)


ericvw:
Should `tslint.json` be updated as well as any dependency on `tslint`?

thmiceli:
Since it may take a while before tslint incorporates my enhancements in a new release (assuming that they will take all of them). I can push here my custom tslint rules. I am not sure if I can override the existing rules that I modified.

thmiceli:
Adding a dependency on the tslint project on github doesn't work because the output of a build is not pushed on github. I could however add a dependency to a fork of tslint that contains everything that is needed to run tslint out-of-the-box.

ericvw:
I would advocate for asking in the PR to `tslint` when they plan to release a new `npm` module.  At least that is what I have done in the past.

ericvw:
PR #110 is a prime candidate why it would be best to directly depend upon the `tslint` changes somehow :)

thmiceli:
I added custom tslint rules and updated makefile so that they get built automatically.

tslint.json
```diff
@@ -1,5 +1,6 @@
 {
     "rules": {
+        "align": [true, "parameters", "arguments", "statements"],
```
ericvw: Let's format this similarly to the other configuration options.  See `typedef` for example.

ericvw:
lgtm - please squash

ericvw:
Also, please update the commit message body to describe in detail the high-level changes that are a part of this change.

Issue #108, opened 2015-02-27
Configure CRL list for node server


ericlu88:
Currently the http server is not supporting CRL. We need to configure the node server to read a CRL. Other consideration including auto-bouncing the server once the CRL change( as part of research in #13 we found out that node server will not update when CRL file changes).

ericlu88:
close #108

PR #110, opened 2015-03-02
Add CRL as part of server https option

ericlu88:
This PR adds CRL as part of server https option. It addresses #108 partially because it does not deal with auto-bouncing the server when CRL changes.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,5 +1,6 @@
 /// <reference path='../typings/tsd.d.ts' />
 
+import events = require('events');
 import fs = require('fs');
 import path = require('path');
 import convict = require('convict');
@@ -17,6 +18,8 @@ export function get(name: string): any
 
     return convictConf.get(name);
 }
+
+export var emitter = new events.EventEmitter();
 
 // Initialization
 ((): void =>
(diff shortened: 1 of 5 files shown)


ericvw:
lgtm

ericlu88:
Updated the PR with auto-bouncing the server when CRL is updated.

lib/config.ts
```diff
@@ -274,9 +283,20 @@ export function get(name: string): any
...
         };
+
+        // Setup file watch for crl changes
+        fs.watch(path.resolve(__dirname, convictConf.get('https.crl')),
+                 (event: string, filename: string): void => {
+                    // Re-read the crl file
+                    otherConf['serverOptions'].httpsServerOptions.crl =
+                        fs.readFileSync(path.resolve(__dirname, convictConf.get('https.crl')));
+                    emitter.emit('change', 'https.crl');  // Signal the server that config changes
+                });
```
thmiceli: Does 'path.resolve' need to be called twice (as an argument to 'watch' and in the callback)? Can't the 'filename' argument to the call back be used instead?

ericlu88: According to [Node document:](http://nodejs.org/api/fs.html#fs_fs_watch_filename_options_listener)

```
Filename Argument#

Providing filename argument in the callback is not supported on every platform (currently it's only supported on Linux and Windows). Even on supported platforms filename is not always guaranteed to be provided. Therefore, don't assume that filename argument is always provided in the callback, and have some fallback logic if it is null.
```

But I guess I can at least cache the value in a variable so we don't need to read it multiple time.

yupinghu:
LGTM

ericvw:
lgtm - please squash.

PR #111, opened 2015-03-03
Remove argument array in websocket

ericlu88:
Noticed a small issue while testing `websocket` that the object being sent over the wire is wrapped in a redundant array. 
This is due to the generated js for `...args` syntax.
Since we are not using anywhere else for multiple arguments, remove the unnecessary argument array.


yupinghu:
LGTM

ericvw:
lgtm

PR #117, opened 2015-03-05
build custom tslint rules when linting test files

ericlu88:
Add `build-rules` for `tslint-test` target so the custom rules will be taken into account for test files also.


ericvw:
lgtm - will merge

PR #121, opened 2015-03-05
Updated 'align' rule to handle arguments in constructor calls.

thmiceli:
I didn't notice initially that that constructor invocation is folded into a 'new expression'.


ericvw:
lgtm - will merge since this is straight forward once TravisCI is happy.

PR #122, opened 2015-03-06
bind this in _.map

ericlu88:
Fix problem in the `blpapi-wrapper.ts` where we didn't bind `this` arg specifically so `this.nextCorrelatorId()` will fail.


yupinghu:
LGTM, sorry for introducing this.

PR #125, opened 2015-03-09
Add boolean option for controling whether to use CRL on server

ericlu88:
This changes add a boolean option `crlEnable` to control whether we want to use CRL on the server. The default option is `false`.

This is a follow-up changes for #108

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -84,7 +84,7 @@ export var emitter = new events.EventEmitter();
             'crl': {
                 doc: 'HTTPS server certificate revocation list',
                 format: String,
-                default: '../keys/hackathon-crl.pem',
+                default: '',
                 arg: 'https-crl'
             }
         },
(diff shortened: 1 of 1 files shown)


ericvw:
Could the boolean be implicit if the CRL file path is provided in the config?  It would reduce the config options to just one path.

ericlu88:
I've changed the default value of `crl` config to be an empty string, and use that as a flag for not using `crl` on the server.

yupinghu:
LGTM, will defer to @ericvw if he has other comments.

ericvw:
Does `null` work instead of empty string?  It would be preferable; if it doesn't work, don't worry about it.

ericlu88:
no I've tried it but `convict` complaint the default value is not a type of `String`

ericvw:
Cool.  lgtm - please squash

ericlu88:
Squashed.

Issue #17, opened 2015-01-13
Provide helper script for generating certificates and keys


ericvw:
It will be useful to have a script used for development that generate server and client-side certificates.

ericvw:
Done in #126

PR #126, opened 2015-03-09
Add gen_keys.sh for generating ssl key and certificate for test server

ericlu88:
This PR adds a bash script for generating ssl ca, server certificate and client certificate. It would be useful for both development and testing purpose.

This addresses #17

### DIFF
--- a/scripts/gen_keys.sh
+++ b/scripts/gen_keys.sh
@@ -1,0 +1,50 @@
+#! /usr/bin/env bash
+
+# WARNING: This script is only intended for development and testing purpose.
+#          DO NOT USE IN PRODUCTION ENVIRONMENT
+# usage: ./GenKey.sh [<path>]
+
+# check if openssl exist
+if command -v openssl >/dev/null; then
+    echo $(openssl version)
+else
+    echo "openssl does not exist"
+    exit 1
+fi
+
+gen_cert() {
+    if [ "$#" -ne 3 ]; then
+        echo "Missing parameter"
+        exit 1
+    fi
+
+    DIR="$1"
+    NAME="$2"
+    SUB="$3"
+    # generate private key
+    openssl genrsa -out "$DIR"/"$NAME"-key.pem 1024
+    # generate csr
+    openssl req -new -key "$DIR"/"$NAME"-key.pem \
+        -out "$DIR"/"$NAME"-csr.pem -text -subj "$SUB"
+    # generate cert
+    openssl x509 -req -in "$DIR"/"$NAME"-csr.pem \
+        -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
+        -set_serial 01 -out "$DIR"/"$NAME"-cert.pem -text
+}
+
+# make directory for certs and keys, if receive path argument
+DIR=${1:-${PWD}}
+if [ "$#" -eq 1 ]; then
+    mkdir -p "$DIR"
+fi
+
+# generate ca private key
+openssl genrsa -out "$DIR"/ca-key.pem 1024
+# generate ca
+openssl req -new -x509 -text -key "$DIR"/ca-key.pem \
+    -out "$DIR"/ca-cert.pem -subj "/CN=BLPAPI_HTTP_TEST_CA/"
+
+# generate server cert
+gen_cert "$DIR" server "/CN=localhost/"
+# generate client cert
+gen_cert "$DIR" client "/CN=BLPAPI_HTTP_TEST_CLIENT/"


ericvw:
I know we the other directories are singular, but many projects pluralize script (i.e., `scripts`).

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
ericvw: `#! /usr/bin/env bash`

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# usage: ./GenKey.sh [<directory_name>]
+
+# get blpapi-http root directory
+DNAME=${1:-temp}
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && mkdir "$DNAME" && cd "$DNAME" && pwd )
```
ericvw: This can be simplified a bit and there is no need to use `cd`.  http://mywiki.wooledge.org/BashFAQ/035

I would recommend having the usage be like this:

./gen_keys.sh <dir_name>, where the directory name is the current working directory if the directory is not specified.

ericvw: Once you process the args you can just:
`mkdir -p <specified-directory>`

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
...
+openssl x509 -req -in "$DIR"/server-csr.pem \
+    -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
+    -set_serial 01 -out "$DIR"/server-cert.pem -text
+
+# generate client cert
+openssl req -new -key "$DIR"/client-key.pem \
+    -out "$DIR"/client-csr.pem -text -subj "/CN=BLPAPI_HTTP_TEST_CLIENT/"
+openssl x509 -req -in "$DIR"/client-csr.pem -extensions usr_cert \
+    -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
+    -set_serial 02 -out "$DIR"/client-cert.pem -text
```
ericvw: Does `-text` work?

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
...
+    -out "$DIR"/ca-cert.pem -subj "/CN=BLPAPI_HTTP_TEST_CA/"
+
+# generate server cert
+openssl req -new -key "$DIR"/server-key.pem \
+    -out "$DIR"/server-csr.pem -text -subj "/CN=localhost/"
+openssl x509 -req -in "$DIR"/server-csr.pem \
+    -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
+    -set_serial 01 -out "$DIR"/server-cert.pem -text
+
+# generate client cert
```
ericvw: The generation and signing of certs could have better re-use if these steps were a function.  http://mywiki.wooledge.org/BashGuide/CompoundCommands#Functions

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# usage: ./GenKey.sh [<directory_name>]
+
```
ericvw: At the top here, I would do a `command` test to ensure that `openssl` is available or fail gracefully.  http://mywiki.wooledge.org/BashFAQ/081

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
...
+openssl req -new -key "$DIR"/server-key.pem \
+    -out "$DIR"/server-csr.pem -text -subj "/CN=localhost/"
+openssl x509 -req -in "$DIR"/server-csr.pem \
+    -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
+    -set_serial 01 -out "$DIR"/server-cert.pem -text
+
+# generate client cert
+openssl req -new -key "$DIR"/client-key.pem \
+    -out "$DIR"/client-csr.pem -text -subj "/CN=BLPAPI_HTTP_TEST_CLIENT/"
+openssl x509 -req -in "$DIR"/client-csr.pem -extensions usr_cert \
```
ericvw: What does `usr_cert` provide?

script/GenKey.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# usage: ./GenKey.sh [<directory_name>]
```
ericvw: Let's rename the script to `gen_keys.sh` and I would also put a big warning comment up here that this script should not be used for generating keys for production usage.

script/GenKey.sh
ericlu88: Like we discovered last week, it works in 0.9.8, and in 1.x.x it print text to stdout. It doesn't hurt to include `-text` options in both cases.

script/GenKey.sh
ericlu88: I tries to add some `X509v3 extensions` elements to the issued certificate. I missed it while testing the ssl issue unintentionally. Will remove it later.

ericvw:
lgtm

ericvw:
Actually, update the commit message to be `gen_keys.sh` to match the file name.  Then lgtm :D

ericlu88:
Will appreciate another eye before we could merge it. @yupinghu @thmiceli

thmiceli:
LGTM

Issue #124, opened 2015-03-07
Subscribe/unsubscribe message events should echo correlations ids


ericvw:
Currently there is no way to distinguish which subscriptions have been successfully subscribed or unsubscribed.

PR #116, opened 2015-03-05
Testing infrastructure

ericlu88:
This PR adds the basic testing infrastructure for http server end-to-end test.

Specifically, it implements a mock `blpapi-node` layer that behaves according to instructions. The instructions can be set beforehand or live via ipc channel(a `socket.io` connection).

For test cases, the routine is to create another node process that host the server(TestHelper.ts has a helper function for this), and let the test cases act as client to hit the server. When the test server boost up, it uses mockery to hijack the dependency to `blpapi` and substitute it with a mocking layer `mock-blpapi.ts`.

This PR also adds test cases for request/response (`request-response.test.ts`) and subscriptions over socket-io ('subscription-socket-io.test.ts').

This is the first bit of issue #18.

PR #127, opened 2015-03-09
Send correlation ids with subscribed and unsubscribed events.

thmiceli:
Also modified examples and changed the unsubscribe event name to 'unsubscribed'.
Issue #124.

### DIFF
--- a/examples/node/MarketDataSubscription.js
+++ b/examples/node/MarketDataSubscription.js
@@ -12,25 +12,26 @@ var opt = {
     key: fs.readFileSync('client.key'),
     ca: fs.readFileSync('bloomberg.crt')
 };
-var NUM = Infinity; // Number of subscription data received before unsubscribe
+var NUM = 2; // Number of subscription data received before unsubscribe
 var counter = 0;
+var unsubscribedCounter = 0;
 
 var socket = io.connect(url, opt);
+
+var subscriptions = [
+    { security: 'AAPL US Equity', correlationId: 0, fields: ['LAST_PRICE'] },
+    { security: 'GOOG US Equity', correlationId: 1, fields: ['LAST_PRICE'] }
+];
 
 socket.on('connected', function () {
     counter = 0;
     console.log('Connected');
-    socket.emit('subscribe',
-                [
-                    { security: 'AAPL US Equity', correlationId: 0, fields: ['LAST_PRICE'] },
-                    { security: 'GOOG US Equity', correlationId: 1, fields: ['LAST_PRICE'] }
-                ]
-    );
+    socket.emit('subscribe', subscriptions);
 });
 
 socket.on('data', function (data) {
     console.log('Data Count: ' + counter++);
-    console.log(data);  
+    console.log(data);
     if (counter === NUM) {
         socket.emit('unsubscribe');
         // Unsubscribe only partial subscription
@@ -42,18 +43,19 @@ socket.on('err', function (data) {
     console.log(data);
 });
 
-socket.on('subscribed', function () {
+socket.on('subscribed', function (correlationIds) {
     console.log('Subscribed');
+    console.log('Correlation Ids: ' + correlationIds);
 });
 
-socket.on('unsubscribed', function () {
+socket.on('unsubscribed', function (correlationIds) {
     console.log('Unsubscribed');
-});
-
-socket.on('unsubscribed all', function () {
-    console.log('Unsubscribed all');
-    socket.disconnect();
-    process.exit();
+    console.log('Correlation Ids: ' + correlationIds);
+    unsubscribedCounter += correlationIds.length;
+    if (unsubscribedCounter === subscriptions.length) {
+        socket.disconnect();
+        process.exit();
+    }
 });
 
 socket.on('disconnect', function() {
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -33,7 +33,19 @@ var AUTH_ACTION_MAP: ActionMap = {
     'authorize': onAuthorize
 };
 
+type Properties = {
+    [index: string]: any;
+}
+
 // PRIVATE FUNCTIONS
+function getSubscriptionsProp(subscriptions: Subscription[]): Properties {
+    var corrIds = subscriptions.map((s: Subscription): number => {
+        return s.correlationId;
+    });
+    var result: Properties = { 'correlationIds': corrIds };
+    return result;
+}
+
 function validatePollId(session: Session, newPollId: number): { isValid: boolean;
                                                                 fetchNewData?: boolean; }
 {
@@ -229,7 +241,7 @@ function onSubscribe(req: Interface.IOurRequest,
                     req.apiSession.activeSubscriptions.set(s.correlationId, s);
                 });
                 req.log.debug('Subscribed');
-                res.sendEnd(0, 'Subscribed');
+                res.sendWhole(0, 'Subscribed', getSubscriptionsProp(subscriptions));
             } else { // Unsubscribe if session already expires
                 req.blpSession.unsubscribe(subscriptions);
                 subscriptions.forEach((s: Subscription): void => {
@@ -310,7 +322,8 @@ function onUnsubscribe(req: Interface.IOurRequest,
         }
 
         res.sendChunk(result);
-        res.sendEnd(0, 'Unsubscribe Successfully');
+        res.sendOtherProp(getSubscriptionsProp(subscriptions));
+        res.sendEnd(0, 'Unsubscribed Successfully');
         return next();
     } catch (err) {
         req.log.error(err, 'Unsubscription error');
@@ -340,7 +353,7 @@ function onAuthorize(req: Interface.IOurRequest,
     req.blpSession.authorizeUser(req.body)
         .then((identity: blpapi.Identity): void => {
             auth.setIdentity(req, identity);
-            res.sendEnd(0, 'OK');
+            res.sendWhole(0, 'OK');
             return next();
         })
         .catch((err: Error): void => {
@@ -368,6 +381,7 @@ export function elevateRequest (req: Interface.IOurRequest,
                                 next: restify.Next): void
 {
     var chunkIndex: number = 0;
+    var otherProperties: Properties = {};
 
     /**
      * This is called each time data is added to the response.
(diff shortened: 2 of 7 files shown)


yupinghu:
FYI: You should eventually write the text "Closes #N" in the commit message before submitting, so that it'll automatically do that. :)

examples/node/MarketDataSubscription.js
```diff
@@ -32,7 +32,7 @@ socket.on('data', function (data) {
     console.log('Data Count: ' + counter++);
     console.log(data);  
     if (counter === NUM) {
-        socket.emit('unsubscribe');
+        socket.emit('unsubscribe', { correlationIds: [0, 1] });
```
ericlu88: This should not be needed here. The intention is that if user send `unsubscribe` event without `correlationIds`, the server will unsubscribe all active subscriptions.

lib/websocket/websocket-handler.ts
```diff
@@ -152,7 +158,7 @@ function setup(socket: Interface.ISocket): void
                         activeSubscriptions.set(s.correlationId, s);
                     });
                     socket.log.debug('Subscribed');
-                    socket.notifySubscribed();
+                    socket.notifySubscribed(getCorrelationIds(subscriptions));
```
ericlu88, line 168: I don't have a strong feeling of this, but since we have the loop of all `subscriptions` @line 157, it might be advantageous to build the array there for performance reason. Same for unsubscribe below.

thmiceli, line 168: I often hesitate myself between reusing an existing loop and creating a new one. In this case I find it clearer to factor the correlation ids extraction into its own function.

examples/node/MarketDataSubscription.js
```diff
@@ -42,16 +42,19 @@ socket.on('err', function (data) {
...
 });
 
-socket.on('unsubscribed', function () {
+socket.on('unsubscribed', function (correlationIds) {
     console.log('Unsubscribed');
+    console.log('Correlation Ids: ' + correlationIds);
 });
 
-socket.on('unsubscribed all', function () {
+socket.on('unsubscribed all', function (correlationIds) {
```
ericvw: Now that we have correlation id's, I wonder if 'unsubcribed all' is necessary to echo back to the client anymore.  @ericlu88, what do you think?

lib/websocket/websocket-handler.ts
ericvw, line 168: I like the separate loop because it is clear on the intent of this function and is not conflated with other logic.  I doubt we will have a high-rate of `subscribe` calls where this would show up as a performance concern.

ericvw:
For the commit message:

```
<subject of commit here up to some reasonable line length (100 chars?)>
<emtpy line>
<body of commit message with each line wrapped at 80 cols - optional>
Closes #1234
```

ericvw:
Let's also try to settle upon if unsubscribe all is necessary now before merging.

examples/node/MarketDataSubscription.js
ericlu88: I think it make sense to get rid of `unsubscribed all` event now. The client should keep track of all correlation ids that then sent to our server, and if they see all of them get unsubscribed, they could decide to disconnect.

ericlu88:
It just occurs to me that we should keep the behavior in sync with long-polling. I've created #128 for changing the long-polling subscription logic. Just FYI.

thmiceli:
I will roll into this PR the changes for #128 unless somebody objects.

examples/node/MarketDataSubscription.js
thmiceli: I agree, `unsubscribed all` is redundant and error prone, the client has to remember that he has to listen to 2 different 'unsubscribed' events.

yupinghu:
+1 to just doing all correlation ids in one change.

ericvw:
+1 for all in one change as well.

ericlu88:
FYI @thmiceli Once the testing-infrastructure PR #116 landed, it might be worth to add these checks into the test cases. Thanks

thmiceli:
I updated the PR with a solution for #128. Also I removed the 'unsubscribed all' event.
I still have to update the new unit tests to take into account the enhancements.

thmiceli:
Look like CI is failing because one of the tests is timing out. Not sure if this is caused by my changes.

thmiceli:
@ericlu88 showed me that the CI failures are caused by the removal of the "unsubscribe all" event. So this is all good and will get fixed with an update of the regression tests.

lib/middleware/request-handler.ts
```diff
@@ -406,11 +412,17 @@ export function elevateRequest (req: Interface.IOurRequest,
         res.write(text + JSON.stringify(data));
     };
 
-    res.sendEnd = (status: string, message: string): void => {
-        var text = util.format('%s%s,%s}',
+    res.sendEnd = (status: string,
```
ericlu88: I think we could reuse `res.sendChunk()` for sending back the `correlationIds` instead of modifying `res.sendEnd()`.
Also, please add the same functionality for `onSubscribe` method.

yupinghu: I agree with Eric's point overall. One thing to note, though: using `sendChunk` will result in slightly different output, specifically `correlationIds` will be in the `data` property of the response, rather than as a top level property. I don't think this really matters, though?

ericlu88: I think as long as we have something like `{data: {correlationIds: [1, 2, 3]}}` we are good.

thmiceli: Actually 'data' is an array, using 'sendChunk' unmodified will result in something like that:
{ data: [ { correlationIds: [1,2] } ], ... }
I can create a 'sendCorrelationIds' function which adds a top property to the response, that would be better than tweaking either 'sendChunk' or 'sendEnd'.

yupinghu: That will actually be kind of complicated, because the functions as currently written assume that you have zero or more calls to `sendChunk` followed immediately by a call to `sendEnd` with no other calls in between writing to the result. We may want to rethink how those functions work if we anticipate more of this sort of thing.

In the short run I might prefer the approach you have here (i.e. augmenting `sendEnd` with final data, although maybe more generically, e.g., supplying both key & value instead of assuming it's `correlationIds`) over trying to make `sendCorrelationIds` work, just for simplicity's sake.

thmiceli: +1

yupinghu: If you go down this route, here's a suggested interface: replace the `correlationIds` param with an `additionalProperties` param of type `{ [index: string]: any }` which gets written to the output. Either flatten them into the top level or, if you want to not have any potential collisions, write it as-is.

thmiceli: @yupinghu I was thinking about something along these lines but I wasn't sure how to express it. Thanks!

ericlu88: Well, `{ data: [ { correlationIds: [1,2] } ], ... }` isn't too bad to me, and it is far more simpler than other proposed approach. Besides, the only case where the `data` array contains more than `correlationIds` is when `unsubscribe` request come in and there are some unpolled buffer data, so we want to send back the last chunk. This should be a relative rare case.

yupinghu: I could go either way personally -- if it's ok to have `correlationIds` embedded in the data, then great, but if that's not optimal, then I'd rather augment `sendEnd` in the short run and maybe have a TODO for improving it later.

ericvw: I would recommend we have a `sendRaw` that does no wrapping/structure of the response and just sends the entire buffer-in full to the client.  Then you can just send the correlation ids that way.

thmiceli: I wouldn't like to have the correlation ids as one of the element of the data array and that they may be intermixed with elements of a different kind (even if that is a rare occurence this case still needs to be handled). It may simplify our code but will complexify the handling of the response.
Extending (optionally) the response with some generic properties is not that complex and will only add an optional parameter to sendEnd.

ericvw: As a client of using WebSockets/SocketIO, I would advocate for only providing the array of correlation ids.

yupinghu: Based on our offline discussion, here's the API I'd propose for this (I'll use the existing names although we may consider renaming them):
- `sendChunk` gets a boolean `isLast` parameter, to decouple the signaling of when chunk data is done from wrapping up the message.
- `sendEnd` stays mostly the same: it adds the core properties (`status` and `message`) to the output which are part of the overall API (i.e., every HTTP response we send has these properties which are used to signal error messages etc. and which I believe we should continue to do for all responses) and flushes the output. The only modification is that it doesn't terminate the data array since `sendChunk` now does that.
- We introduce a new `sendOtherProperty` function (I know, that name sucks, but just for discussion) which writes a property to the output. This function either has to be constrained to be called before or after all `sendChunk` calls, or we can add buffering for when it's called in the middle and write them to the response during the last `sendChunk`.
- We also add a `sendWholeMessage` function, which may reuse `sendOtherProperty` and/or `sendEnd` internally.

Which leads to an API that says you have two choices:

1) Use `sendWholeMessage` and nothing else.
2) Use any number of `sendOther` and `sendChunk` calls, where the last `sendChunk` needs to specify that it is, followed by `sendEnd`.

Thoughts?

thmiceli:
As discussed I added the 'sendOtherProp' function that extends the response with whatever properties it gets passed.

lib/middleware/request-handler.ts
```diff
@@ -406,11 +421,25 @@ export function elevateRequest (req: Interface.IOurRequest,
...
+        _.assign(otherProperties, properties);
+    };
+
     res.sendEnd = (status: string, message: string): void => {
-        var text = util.format('%s%s,%s}',
+        var text = util.format('%s%s,%s',
                                prepareResponseText(true),
                                stringifyPair('status', status),
                                stringifyPair('message', message || ''));
+        if (otherProperties) {
```
ericlu88: `otherProperties` will always yield true now.

thmiceli:
I added the test because typescript complains about a for loop on an object properties without a check. I'll use _.forEach then.

lib/middleware/request-handler.ts
```diff
@@ -406,11 +421,25 @@ export function elevateRequest (req: Interface.IOurRequest,
...
+            for (var property in otherProperties) {
```
ericlu88: this can be simplified by using `_.forOwn()`

thmiceli:
I updated the regression tests. @ericlu88, could you please take a look?

thmiceli:
I addressed the latest comments and updated the PR.

ericlu88:
Based on our last discussion, I think having a _sendAll_ function would be pretty helpful. The API imo looks like this(The function name of course is not settled): 

``` typescript
sendAll(data?: any,
        properties?: { [index: string]: any; },
        status:any,
        message: string): void
```

As for the implementation, I don't think re-use the existing `sendChunk` and `sendEnd` is a good idea since they may still send data in different chunk, which is not the user would expect from the API. I would recommend wrapping around restify's `res.send()` function.

thmiceli:
Do we want the additional function in this PR? Personally I don't mind.
Could we just call it 'send'?

ericlu88:
I think it make sense to have it in this PR so that we could avoid calls like `res.sendChunk(); res. sendOtherProp(); res.sendEnd()` in 3 lines.

I am ok with calling it `res.send()`. The only caveat is that you have to monkey patch the function since restify already defined it.

@yupinghu brings up a good point of keeping the code that structure our response centralized. We may need to take that into consideration when we implement it.

thmiceli:
I added the function to send an entire response in one shot. I called it 'sendWhole'.

yupinghu:
Last commit LGTM, someone else take a quick look as well?

ericlu88:
lgtm except some nit parameter alignment issue.

thmiceli:
I addressed the last comments from @ericlu88 and squashed the commits.

PR #129, opened 2015-03-10
Quick fix for requests without identities.

yupinghu:
When I switched from undefined -> null I didn't realize
this wouldn't work because blpapi-node expects an object or
undefined.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -277,6 +277,8 @@ export class Session extends events.EventEmitter {
             if (isAuthRequest) {
                 this.session.authorizeUser(request, correlatorId);
             } else {
+                // TODO: blpapi-node doesn't accept null; remove this when that's fixed.
+                identity = identity || undefined;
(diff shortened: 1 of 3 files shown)


ericvw:
Please rebase this PR due to the history-rewrite.

ericlu88:
Please also update the version of `blpapi` in  `package.json` so when we run `make` will automatically pick up the latest dependency.

lib/blpapi-wrapper.ts
```diff
@@ -277,6 +277,10 @@ export class Session extends events.EventEmitter {
             if (isAuthRequest) {
                 this.session.authorizeUser(request, correlatorId);
             } else {
+                if (identity === null) {
```
ericvw: nit:  Alternatively: `identity = identity || undefined;`

yupinghu:
Rebased and updated; ready for another look.

ericlu88:
LGTM.

ericvw:
lgtm.  Will merge once TravisCI is happy.

PR #134, opened 2015-03-12
Add test cases for long-poll subscription over https and fix bugs

ericlu88:
This PR adds test cases for long-poll subscriptions over https. It also fix bugs that revealed by the test cases. Specifically, they are:
- Fix problem in `blpapi-wrapper.ts` where `unsubscribe` used to only unlisten _MarketDataEvents_
- Fix problem in `blp-session.ts` by removing the max number of listen on `blpSession` (by default is 10). This is needed because every created `apiSession` shares the same `blpSession`, and every `apiSession` is listening on `SessionTerminated` to clear the state.
- Fix problem in `request-handler.ts` where `unsubscribe` only send unstructured data(no `correlationId`, `missed_ticks` sent).
- Fix problem in `request-handler.ts` where validating subscription options will throw exception before the promise get created.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -295,6 +295,7 @@ export class Session extends events.EventEmitter {
         this.session.once('SessionTerminated', this.sessionTerminatedHandler.bind(this));
         log('Session created');
         trace(opts);
+        this.setMaxListeners(0);  // Remove max listener limit
     }
 
     // MANIPULATORS
@@ -503,7 +504,14 @@ export class Session extends events.EventEmitter {
                 this.subscriptions[cid].emit('end');
                 delete this.subscriptions[cid];
             });
-            this.unlisten('MarketDataEvents', cid);
+            // unlisten for events
+            var service = securityToService(this.subscriptions[cid].security);
+            assert(service in SERVICE_TO_SUBSCRIPTION_EVENTS_MAP,
+                   util.format('Service, %s, not handled', service));
+            var events = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[service];
+            _.forEach(events, (e: string): void => {
+                this.unlisten(e, cid);
+            });
         });
     }
 }
--- a/lib/middleware/request-handler.ts
+++ b/lib/middleware/request-handler.ts
@@ -23,6 +23,14 @@ type ActionMap = {
                       next: restify.Next) => any;
 };
 
+// Type used for subscription input from request body
+type SubscriptionOption = {
+    correlationId: number;
+    security: string;
+    fields: string[];
+    options?: any;
+}
+
 var SUBSCRIPTION_ACTION_MAP: ActionMap = {
     'start': onSubscribe,
     'stop':  onUnsubscribe
@@ -90,17 +98,15 @@ function validatePollId(session: Session, newPollId: number): { isValid: boolean
     }
 }
 
-function startAllNewBuffers(subscriptionStore: Map<Subscription>): Object[]
+function startAllNewBuffers(subscriptions: Subscription[]): Object[]
 {
     var buffers: Object[] = [];
     // Check if we have new data for any subscriptions
-    var hasNewData: boolean = false;
-    subscriptionStore.forEach((sub: Subscription): boolean => {
-        hasNewData = !sub.buffer.isEmpty();
-        return sub.buffer.isEmpty();
+    var hasNewData: boolean = _.some(subscriptions, (sub: Subscription): boolean => {
+        return !sub.buffer.isEmpty();
     });
     if (hasNewData) { // Start a new buffer for every subscription and return back the new data
-        subscriptionStore.forEach((sub: Subscription): boolean => {
+        subscriptions.forEach((sub: Subscription): void => {
             var buff: Interface.IBufferedData<Object> = sub.buffer.startNewBuffer();
             if (buff.buffer.length) {
                 buffers.push({
@@ -109,7 +115,6 @@ function startAllNewBuffers(subscriptionStore: Map<Subscription>): Object[]
                     'missed_ticks': buff.overflow
                 });
             }
-            return true;
         });
     }
     return buffers;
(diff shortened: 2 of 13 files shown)


lib/blpapi-wrapper.ts
```diff
@@ -460,30 +465,36 @@ export class Session extends events.EventEmitter {
 
         log('Unsubscribing: ' + JSON.stringify(subscriptions));
 
-        var cids: number[] = [];
-        _.forEach(subscriptions, (s: Subscription, i: number): void => {
+        var caes = _.map(subscriptions, (s: Subscription): CidAndEvents => {
```
ericvw: Let's keep this as `number[]` and perform the look-up/unlisten of the events in the `_.forEach()` below?  It's less change to`unsubscribe` and the same end-result is achieved with no performance impact.

scripts/gen_client_cert.sh
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env bash
```
ericvw: Since this is near duplicated of gen_keys.sh, I would recommend making this a sourcable bit of bash code that can be used by gen_keys.sh.  See http://mywiki.wooledge.org/BashGuide/Sourcing

test/lib/test-helper.ts
```diff
@@ -6,59 +6,101 @@ import Promise = require('bluebird');
...
 
 var log = debug('child_server:stdout');
+// CONST
+var SESSION_EXPIRATION: number = 3;
+var MAXBUFFERSIZE: number = 3;
+var POLL_TIMEOUT: number = 1000;
+
+function createCertificate(): Promise<any>
+{
+    return (Promise.promisify(child.execFile))('./scripts/gen_keys.sh', ['./temp']);
```
ericvw: Since this is test code, if there is an API, I would just make this synchronous.

test/lib/test-helper.ts
```diff
@@ -6,59 +6,101 @@ import Promise = require('bluebird');
...
+}
+
+function removeCertificate(): Promise<any>
```
ericvw: Since this is test code, if there is an API, I would just make this synchronous.

lib/middleware/blp-session.ts
```diff
@@ -20,6 +20,7 @@ function createSession(sessionId: string): Promise<blpapi.Session>
 {
     if (!(sessionId in SESSION_STORE)) {
         var s = new blpapi.Session(DEFAULT_SESSION_OPTIONS);
+        s.setMaxListeners(0);   // Remove max listener limit for blpSession
```
ericvw: Let's move to using `.defaultMaxListeners` in `blpapi-wrapper.ts`, which can be a 'class wide' property that is the default for all new EventEmitter instances.

lib/middleware/request-handler.ts
```diff
@@ -163,66 +163,76 @@ function onSubscribe(req: Interface.IOurRequest,
...
     var subscriptions: Subscription[] = [];
-    ((): Promise<void> => {
-        // Check if req body is valid
-        if (!_.isArray(req.body) ||
-            !req.body.length) {
-            throw new Error('Invalid subscription request body.');
+    if (!_.isArray(req.body) || !req.body.length) {
+        return next(new restify.BadRequestError('Invalid subscription request body.'));
+    }
+    _.forEach(req.body, (s: {'correlationId': number;
```
ericvw: `_.every` is probably more appropriate for setting `isValid`. https://lodash.com/docs#every

lib/middleware/request-handler.ts
```diff
@@ -163,66 +163,76 @@ function onSubscribe(req: Interface.IOurRequest,
                      res: Interface.IOurResponse,
                      next: restify.Next): void
 {
+    // Validate input options
```
ericvw, line 183: Generally, it may be a better approach to do the validation in one past and map down the valid subscriptions to subscribe in another pass.  It separates the steps clearly for subscriptions instead of conflating them in one massive _.forEach that may return early and has unclear side-effects.

test/lib/test-helper.ts
```diff
@@ -6,59 +6,101 @@ import Promise = require('bluebird');
...
+        serverOptions.push('--https-enable=true');
+        serverOptions.push('--https-ca=../temp/ca-cert.pem');
+        serverOptions.push('--https-cert=../temp/server-cert.pem');
+        serverOptions.push('--https-key=../temp/server-key.pem');
+        serverOptions.push('--logging-clientDetail=true');
+        serverOptions.push('--session-expiration=' + SESSION_EXPIRATION);
+        serverOptions.push('--longpoll-maxbuffersize=' + MAXBUFFERSIZE);
+        serverOptions.push('--longpoll-polltimeout=' + POLL_TIMEOUT);
+    }
+    var p: Promise<any> = https ? createCertificate() : Promise.resolve();
```
ericvw, line 55: I would make this synchronous if possible, it just simplifies everything and it is just for testing.

test/lib/test-helper.ts
```diff
@@ -6,59 +6,101 @@ import Promise = require('bluebird');
...
             });
-        });
-    }));
+        }));
+    });
+}
+
+export function stopServer(server: IOurChildProcess): Promise<void>
+{
+    var p: Promise<any> = server.isHTTPS ? removeCertificate() : Promise.resolve();
```
ericvw: There may be a more generic way to deal with clean-up that doesn't require `if`/`switch` type behavior that executes a queued up run-list or array of callbacks to execute for cleanup.

lib/middleware/blp-session.ts
ericlu88: Setting `EventEmitter.defaultMaxListeners` property, according to Node's doc, will affect all EventEmitter instances, not only limited to `blpapi.Session`.

Instead, I will move the `setMaxListener` in the `blpapi.Session` constructor so it will be called for all instance of it.

ericlu88:
Unfortunately `child_process.execSync` is not supported in node 0.10. We have to stick with the async version until we fully convert everything to 0.12.

lib/middleware/request-handler.ts
```diff
@@ -78,17 +86,15 @@ function validatePollId(session: Session, newPollId: number): { isValid: boolean
...
+function startAllNewBuffers(subscriptions: Subscription[]): Object[]
 {
     var buffers: Object[] = [];
     // Check if we have new data for any subscriptions
-    var hasNewData: boolean = false;
-    subscriptionStore.forEach((sub: Subscription): boolean => {
-        hasNewData = !sub.buffer.isEmpty();
+    var isAllEmpty: boolean = _.every(subscriptions, (sub: Subscription): boolean => {
         return sub.buffer.isEmpty();
     });
```
thmiceli, line 107: nit: you could use _.some and return !sub.buffer.isEmpty(). This way you will iterate until the first non empty buffer and it is actually the 'all buffers are not empty' state that is interesting to you, so you won't need to negate the result.

ericlu88, line 107: Good catch. using `_.some` is more fluent in logic.

lib/blpapi-wrapper.ts
```diff
@@ -307,6 +307,7 @@ export class Session extends events.EventEmitter {
         this.session.once('SessionTerminated', this.sessionTerminatedHandler.bind(this));
         log('Session created');
         trace(opts);
+        this.setMaxListeners(0);  // Remove max listener limit for blpSession
```
ericvw: `s/ for blpSession//`

lib/blpapi-wrapper.ts
```diff
@@ -470,7 +471,15 @@ export class Session extends events.EventEmitter {
             if (undefined === cid) {
                 throw new Error('Subscription not found at index ' + i);
             } else {
-                cids.push(_.parseInt(cid));
+                var id: number = _.parseInt(cid);
```
ericvw: nit: type definition is not necessary

lib/blpapi-wrapper.ts
```diff
@@ -470,7 +471,15 @@ export class Session extends events.EventEmitter {
...
+                cids.push(id);
+                // unlisten for events
+                _.forEach(SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[
+                              getServiceForSecurity(this.subscriptions[cid].security)
```
ericvw: nit: indent is a bit weird here.  It may be better to capture in a variable what is being passed into `_.forEach`; also it will make this more readable.

scripts/gen_keys.sh
```diff
@@ -12,25 +12,8 @@ else
...
-    # generate csr
-    openssl req -new -key "$DIR"/"$NAME"-key.pem \
-        -out "$DIR"/"$NAME"-csr.pem -text -subj "$SUB"
-    # generate cert
-    openssl x509 -req -in "$DIR"/"$NAME"-csr.pem \
-        -CA "$DIR"/ca-cert.pem -CAkey "$DIR"/ca-key.pem \
-        -set_serial 01 -out "$DIR"/"$NAME"-cert.pem -text
-}
+# source the gen_cert function
+. $( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/fun_gen_cert.sh
```
ericvw: `"$(dirname "$0")/fun_gen_cert.sh"`

lib/websocket/websocket-handler.ts
```diff
@@ -86,48 +95,48 @@ function setup(socket: Interface.ISocket): void
...
                 return false;
             }
+            return true;
+        });
+        if (!isValid) {
+            socket.log.debug(errMessage);
+            socket.sendError(errMessage);
+            return;
+        }
+        if (data.length !== _(data).pluck('correlationId').uniq().value().length) {
```
thmiceli, line 120: Do you need to call 'value()' on the array since you just need the length?

test/lib/test-helper.ts
```diff
@@ -1,64 +1,104 @@
...
-                    'ipc': socket
+        return (new Promise<IOurSocket>((resolve: (result: IOurSocket) => void,
+                                         reject: (error: Error) => void): void => {
+            server.on('message', (m: any): void => {
+                // Connect the ipc socket once test server is ready
+                var socket: any = io.connect('http://localhost:' + testport);
+                socket.emitAcknowledge = (event: string, arg?: any): Promise<void> => {
+                    return (new Promise<void>((resolve: () => void,
+                                               reject: (error: Error) => void): void => {
+                        var argument = (2 === arguments.length) ? arg : null;
```
thmiceli: I believe that this would be equivalent and independent from the position of 'arg' in the parameter list: 
var argument = arg || null;

lib/websocket/websocket-handler.ts
ericlu88, line 120: The chaining require me to call value() before I can access the array length.

lib/blpapi-wrapper.ts
```diff
@@ -492,7 +493,12 @@ export class Session extends events.EventEmitter {
                 this.subscriptions[cid].emit('end');
                 delete this.subscriptions[cid];
             });
-            this.unlisten('MarketDataEvents', cid);
+            // unlisten for events
+            var service = securityToService(this.subscriptions[cid].security);
+            var events = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[service];
```
ericvw, line 511: Let's add an `assert` here similar to subscribe?

scripts/gen_client_cert.sh
```diff
@@ -0,0 +1,29 @@
...
+        echo "must specify path"
+        exit 1
+    fi
+
+    local -r dir="$1"
+
+    # remove existing client key, csr and cert
+    rm "$dir/client-key.pem"
+    rm "$dir/client-csr.pem"
+    rm "$dir/client-cert.pem"
```
ericvw: Removal shouldn't be necessary since they will just get overridden.

ericlu88:
Rebased and pushed.

ericlu88:
Experience the same issue about `request` module as #151. Has to bound the dependency to 2.53.0.

PR #138, opened 2015-03-17
only listen to subscription events upon successful `subscribe()`

### LINKED ISSUE #115: blpapi-wrapper: Properly deal with errors for all APIs
We should be careful about cleaning up state of the blpapi-wrapper Session when errors occur.

ericvw:
This change primarily addresses the an issue of listening to correlation
id/event names that may have failed when the underlying
`subscribe()` call is made.  Basically, listening of correlation
id/event names happens after the underlying `subscribe()` call is
successful.

The logic has also been simplified for the sake of some additional O(N)
performance.  Once the code is refactored to utilize ES6 Map, the
performance concern will eliminated.

Closes #115

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -113,7 +113,7 @@ function isObjectEmpty(obj: Object): boolean {
     return (0 === Object.getOwnPropertyNames(obj).length);
 }
 
-function getServiceForSecurity(security: string): string {
+function securityToService(security: string): string {
     var serviceRegex = /^\/\/blp\/[a-z]+/;
     var match = serviceRegex.exec(security);
     // XXX: note that we shoud probably capture what the default service is to use when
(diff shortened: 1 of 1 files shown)


lib/blpapi-wrapper.ts
```diff
@@ -445,13 +421,29 @@ export class Session extends events.EventEmitter {
...
+                       util.format('Service, %s, not handled', uri));
+                var events = SERVICE_TO_SUBSCRIPTION_EVENTS_MAP[uri];
+                events.forEach((event: string): void => {
+                    log('listening on event: ' + event + ' for cid: ' + cid);
+                    this.listen(event, cid, (m: any): void => {
+                        userSubscription.emit('data', m.data, s);
+                    });
+                });
+            });
         }).nodeify(cb);
```
ericlu88, line 453: We'll need a `catch` block here to clean up cached `this.subscription[cid]` when `session.subscribe` throw.

ericvw:
Updated PR with `catch`.

ericlu88:
LGTM.

lib/blpapi-wrapper.ts
```diff
@@ -122,15 +122,10 @@ function getServiceForSecurity(security: string): string {
...
-    subscription: Subscription;
-    serviceUri: string;
-}
-
-function getUniqueServices(sass: SubscriptionAndService[]): string[] {
-    return _(sass).map((subscriptionAndService: SubscriptionAndService): string => {
-        return subscriptionAndService.serviceUri;
-    }).uniq().valueOf();
+function subscriptionsToServices(subscriptions: Subscription[]): string[] {
+    return _.chain(subscriptions) .map( (s: Subscription): string => {
```
yupinghu: Were the spaces before `.map` and `(s: Subscription)` intentional?

ericvw:
Addressed whitespace issue.

PR #146, opened 2015-03-18
add support for shellcheck for shell script static analysis

ericvw:
This will help keep bash scripts in check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,12 @@ SRCS_TEST_TS :=
 SRCS_TEST_JS :=
 SRCS_TSLINT_RULES_TS :=
 SRCS_TSLINT_RULES_JS :=
+SRCS_SHELL :=
 
 # known directories within the project
 TEST_DIR := test
 TSLINT_RULES_DIR := tslint-rules
+SCRIPTS_DIR = scripts
 
 # glob all source TypeScript files
 SRCS_TS += index.ts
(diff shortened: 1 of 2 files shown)


ericvw:
Looks like there is more work to getting this working with TravisCI.

ericvw:
Build failure is now due to #152.

ericvw:
Updated PR due to #152 now being resolved.

ericlu88:
LGTM

Makefile
```diff
@@ -77,6 +87,11 @@ clean:
 	@$(RM) $(SRCS_TSLINT_RULES_JS)
 
 
+$(SHELLCHECK_TARGET): $(SRCS_SHELL)
+	@$(RM) $(SHELLCHECK_TARGET)
+	@$(SHELLCHECK) $(SRCS_SHELL)
+	@$(TOUCH) $(SHELLCHECK_TARGET)
```
thmiceli, line 93: Why is the 'touch' needed?

ericvw, line 93: Since `shellcheck` doesn't produce any output, this is the target that gets generated on success.  This way, we only re-run the target if the shell files are newer than the dummy target.

thmiceli, line 93: I see. Since the target is fake it would be helpful to have it identified as such. example SHELLCHECK_FAKE_TARGET.

yupinghu:
LGTM other than the clarification that @thmiceli requested.

PR #147, opened 2015-03-18
Add support for AuthenticationMode=APPLICATION_ONLY

yupinghu:
This change adds a config option for api.authenticationAppName.
When this is set, sessions are created with authentication options
set to APPLICATION_ONLY mode with the specified app name, and
the default Identity for the session is authorized appropriately.
Other authentication modes are not yet supported.

This change also has a small refactor of doRequest to support
the third kind of request -- rather than passing isAuthRequest,
we now pass the actual function to perform the request.

### DIFF
--- a/lib/blpapi-wrapper.ts
+++ b/lib/blpapi-wrapper.ts
@@ -257,9 +257,8 @@ export class Session extends events.EventEmitter {
     private doRequest(uri: string,
                       requestName: string,
                       request: any,
-                      callback: IRequestCallback,
-                      isAuthRequest: boolean,
-                      identity?: blpapi.Identity): void
+                      execute: (correlatorId: number) => void,
+                      callback: IRequestCallback): void
     {
         this.validateSession();
 
@@ -269,13 +268,7 @@ export class Session extends events.EventEmitter {
         this.openService(uri).then((): void => {
             log(util.format('Request: %s|%d', requestName, correlatorId));
             trace(request);
-            if (isAuthRequest) {
-                this.session.authorizeUser(request, correlatorId);
-            } else {
-                // TODO: blpapi-node doesn't accept null; remove this when that's fixed.
-                identity = identity || undefined;
-                this.session.request(uri, requestName, request, correlatorId, identity);
-            }
+            execute(correlatorId);
             assert(requestName in REQUEST_TO_RESPONSE_MAP,
                    util.format('Request, %s, not handled', requestName));
             this.listen(REQUEST_TO_RESPONSE_MAP[requestName],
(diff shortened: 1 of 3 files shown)


ericlu88:
Test to see if @ericvw receive email notification.

lib/blpapi-wrapper.ts
```diff
@@ -257,9 +257,8 @@ export class Session extends events.EventEmitter {
     private doRequest(uri: string,
                       requestName: string,
                       request: any,
-                      callback: IRequestCallback,
-                      isAuthRequest: boolean,
-                      identity?: blpapi.Identity): void
+                      perform: (correlatorId: number) =>  void,
```
ericvw: nit: two spaces in `=>  void` collapse to one

lib/blpapi-wrapper.ts
```diff
@@ -376,8 +396,7 @@ export class Session extends events.EventEmitter {
             }
             var uri = '//blp/apiauth';
             var requestName = 'AuthorizationRequest';
-            var isAuthRequest = true;
-            this.doRequest(uri, requestName, request, callback, isAuthRequest);
+            this.doRequest(uri, requestName, request, performAuthorizeUser, callback);
```
ericvw: You could `this.session.authorizeUser.bind(this.session, request)`.  Then the `performAuthorizeUser` may be removed.

lib/blpapi-wrapper.ts
ericvw: `perform` is a bit ambiguous.  Maybe it is that `doRequest` isn't really doing the request, `perform` is.  I am trying to think of a clearer nomenclature.

ericvw:
`authorize` and `authorizeUser` are quite similar - the only difference is how the token is received.  Since there is some duplication of logic in the JS and binding layer, it seems like exposing the `generateToken` API in the binding layer and, likewise, in the JS layer seems like the best approach.  Then only one API on the wrapper (i.e., `authorizeUser` => `authorize`) is needed.

yupinghu:
Made some changes for the easy feedback to respond to. :) I also wanted to go ahead and config options for all the auth options in the API, to minimize the number of changes users have to make to their configs in the future.

I agree about the poor naming of `doRequest` and `perform`. I'll think about it some more.

Re: `authorize` vs `authorizeUser` -- yes, they should be unified/simplified at some point. I think there's generally cleanup to do in blpapi-node in this area that I'll tackle in the future.

thmiceli:
LGTM

lib/config.ts
```diff
@@ -301,9 +316,35 @@ export var emitter = new events.EventEmitter();
...
 
     // BLPAPI Session options
-    otherConf['sessionOptions'] = {
+    var blpapiSessionOptions: any = {
         serverHost: convictConf.get('api.host'),
         serverPort: convictConf.get('api.port')
     };
 
+    var authenticateOnStartup = false;
+    var authenticationOptions: string;
```
ericvw: nit: initialize to `null` to indicate explicit intent.

lib/config.ts
```diff
@@ -301,9 +316,35 @@ export var emitter = new events.EventEmitter();
...
+        // TODO: Add support for USER_AND_APPLICATION
+        authenticationOptions = util.format('AuthenticationMode=%s;' +
+                                            'ApplicationAuthenticationType=APPNAME_AND_KEY;' +
+                                            'ApplicationName=%s',
+                                            mode,
+                                            appName);
+    }
+    // TODO: Add support for user-only auth modes.
+
+    if (undefined !== authenticationOptions) {
```
ericvw: Updated to be `null` here too.

lib/blpapi-wrapper.ts
```diff
@@ -257,9 +257,8 @@ export class Session extends events.EventEmitter {
     private doRequest(uri: string,
                       requestName: string,
                       request: any,
```
ericlu88, line 259: I'll advocate remove the `request` para since it becomes irrelevant with the `doRequest` method.

lib/middleware/blp-session.ts
```diff
@@ -19,8 +19,15 @@ var LOGGER: bunyan.Logger = bunyan.createLogger(conf.get('loggerOptions'));
 function createSession(sessionId: string): Promise<blpapi.Session>
 {
     if (!(sessionId in SESSION_STORE)) {
-        var s = new blpapi.Session(DEFAULT_SESSION_OPTIONS);
+        var sessionOptions = DEFAULT_SESSION_OPTIONS;
+        var s = new blpapi.Session(sessionOptions.blpapiSessionOptions);
         var p = s.start()
+            .then((): Promise<void> => {
```
ericlu88: Another syntax to save one tick if not authorizing on startup.

``` typescript
if (sessionOptions.authorizeOnStartup) {
    p = p.then(s.authorize);
}
```

ericlu88:
LGTM.

ericvw:
lgtm - waiting on travisci.

PR #149, opened 2015-03-19
Updated dependency on blpapi-node.

### LINKED ISSUE #79: blpapi-wrapper should return different error types

thmiceli:
Closes #79

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Bloomberg Open API HTTP Proxy",
   "main": "server.js",
   "dependencies": {
-    "blpapi": "^1.1.0",
+    "blpapi": "^1.2.0",
     "bluebird": "^2.9.13",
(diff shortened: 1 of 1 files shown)


ericvw:
lgtm - awaiting on TravisCI to merge.

thmiceli:
Looking at the CI failure...

thmiceli:
I don't understand how my change could have made the test fail. I cannot reproduce the failure locally either (tried it with nodejs 0.10). It could be an intermittent failure unrelated with this change.

ericvw:
TravisCI is happy now.  Going to go ahead with this and keep an eye out for the testing failures.

PR #151, opened 2015-03-24
Remove dependency from interface.ts

### LINKED ISSUE #150: Fix for a circular reference issue
Right now there's a circular reference path in our code base:

```
lib/interface.ts->
lib/apisession/session.ts->
lib/subscription/subscription.ts->
lib/util/historical-buffer-manager.ts->
lib/interface.ts
```

Will have to remove this circular reference problem.

ericlu88:
This PR changes the dependencies a bit so `interface.ts` becomes leaf node. It also tries to use interfaces as type definition rather than the concrete implementation. The following graph shows the dependencies after this change. This PR closes #150 .

test

### DIFF
--- a/lib/apisession/session.ts
+++ b/lib/apisession/session.ts
@@ -2,25 +2,27 @@
 
 import bunyan = require('bunyan');
 import BAPI = require('../blpapi-wrapper');
-import Subscription = require('../subscription/subscription');
 import Map = require('../util/map');
 import conf = require('../config');
+import interfaces = require('../interface');
 
 function curSeconds(): number
 {
     return process.hrtime()[0];
 }
 
+type ISubscription = interfaces.ISubscription;
+
 export = Session;
 
-class Session {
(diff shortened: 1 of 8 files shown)


ericlu88:
`request` release 2.54.0 yesterday. However this change breaks `promisify` on request module as several people pointing out already @https://github.com/request/request/issues/1506. I've downgraded the dependency to `request` module back to 2.53.0.

ericvw:
Let's back out the stuff to have `interface.ts` depend upon `blpapi-wrapper.ts`, so that when the API migrates into the `blpapi-node` project, there is less churn in the future.

ericlu88:
@ericvw Agreed. I've undo some of the changes so that the `blpapi-wrapper.ts` is now at the leaf. This will ease the migration when that happens.

test

ericlu88:
Rebase & pushed.

ericvw:
lgtm

Issue #152, opened 2015-04-03
TypeScript 1.5.0-alpha breaks TravisCI build for node v0.10


ericvw:
#151 passed TravisCI in [build #451](http://travis-ci.org/bloomberg/blpapi-http/builds/56134184).  However, after approval and merging into `develop`, TravisCI is complaining for [build #451](https://travis-ci.org/bloomberg/blpapi-http/builds/56435485).

I pushed 70611a7d90ded6bf14f0ca7770cc47c2075aa6a1 directly to `develop` to make this more transparent when viewing the rendered `README.md`.

It appears that only node `v0.10` is picking up TypeScript `1.5.0-alpha` and node `v0.12` and `iojs` is not.

ericvw:
[node-semver](https://github.com/npm/node-semver) documentation for reference.

We are using `"typescript": "^1.4.1"`, which translates to `>= 1.4.1, < 2.0.0`.  However, it doesn't explain the difference between TypeScript version picked up by the different versions of node.

ericvw:
https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes

It looks like TypeScript doesn't adhere to strict semantic versioning - ugh.  It looks like we should use tilde (i.e., `~`) instead of the caret (i.e., `^`).

Investigation is still needed to explain the difference in TravisCI builds - it appears to be related to npm version.

| node version | npm version |
| --- | --- |
| node 0.10.38 | 1.4.28 |
| node 0.12.2 | 2.7.4 |
| iojs 1.6.3 | 2.7.4 |

I bet the semver scheme definition changed between 1.x and 2.x.

ericvw:
As a side-note, we should be aware that TypeScript 1.5 will have breaking changes when we decide to upgrade.

ericvw:
Failure is:

```
lib/server.ts(49,16): error TS2345: Argument of type 'RequestHandler' is not assignable to parameter of type 'RequestHandler[]'.
  Property 'push' is missing in type 'RequestHandler'.
typings/express-validator/express-validator.d.ts(172,3): error TS2309: An export assignment cannot be used in a module with other exported elements.
make: *** [index.js] Error 1
```

However, this could also be a bug in TypeScript `1.5.0-alpha`...

ericvw:
Created #153 to constrain TypeScript to to `1.4.x` versions - TravisCI is happy with that change for now.

ericvw:
[...]

PR #153, opened 2015-04-03
Constrain TypeScript version to allow only for patch-level changes

ericvw:
Due to issue #152, his constrains TypeScript to >= 1.4.1, < 1.5.0.  It
turns out that from npm v1 => v2, there was a change in `semver` in how
caret (i.e., ^) handles preleases that changed.  Since it appears that
TypeScript has breaking changes between minor versions, we need to be
more restrictive on version constraints for `typescript`.


ericlu88:
LGTM.

